PR #13, opened 2018-05-22
OCLOMRS-16: Fix Login UI, Add preloader and logout functionality

hadijahkyampeire:
# JIRA TICKET NAME:
[OCLOMRS-16: Fix Login UI and add preloader](https://issues.openmrs.org/browse/OCLOMRS-16)

# Summary:
This PR fixes the Login component as advised during the demo: 
- the Login component is rendered as the index page.
- Adds a loader component that will show the user the progress when the request is being made.
- Implements logout

### DIFF
--- a/src/Actions/AuthActions.js
+++ b/src/Actions/AuthActions.js
@@ -1,27 +1,35 @@
-import axios from "axios";
-import { Login, LoginFailed } from "../ActionCreators/AuthActionCreators";
+import axios from 'axios';
+import {
+  Login,
+  LoginFailed,
+  LoginStarted,
+  Logout,
+} from '../ActionCreators/AuthActionCreators';
 
-const BASE_URL = "https://api.openconceptlab.org/";
+const BASE_URL = 'https://api.openconceptlab.org/';
 
-// login action
-const LoginAction = ({ username, token }) => {
-  return async dispatch => {
-    try {
-      const url = `${BASE_URL}/users/${username}/`,
-        headers = { Authorization: `Token ${token}` };
-      // dispatch login sent
-      const response = await axios.post(url, null, { headers });
-      dispatch(Login(response));
-      localStorage.setItem("token", `Token ${token}`);
-      localStorage.setItem("username", `${username}`);
-    } catch (error) {
-      if (error.response) {
-        dispatch(LoginFailed(error.response.data.detail));
-      } else if (error.request) {
-        dispatch(LoginFailed("Request can't be made"));
-      }
+const LoginAction = ({ username, token }) => async (dispatch) => {
+  try {
+    const url = `${BASE_URL}/users/${username}/`;
+    const headers = { Authorization: `Token ${token}` };
+    dispatch(LoginStarted());
+    const response = await axios.post(url, null, { headers });
+    dispatch(Login(response));
+    localStorage.setItem('token', `Token ${token}`);
+    localStorage.setItem('username', `${username}`);
+  } catch (error) {
+    if (error.response) {
+      dispatch(LoginFailed(error.response.data.detail));
+    } else if (error.request) {
+      dispatch(LoginFailed("Request can't be made"));
     }
-  };
+  }
 };
 
-export { LoginAction };
+const LogoutAction = () => (dispatch) => {
+  dispatch(LoginStarted());
+  dispatch(Logout());
+  localStorage.clear();
+};
+
+export { LoginAction, LogoutAction };
--- a/src/App.css
+++ b/src/App.css
@@ -32,6 +32,7 @@
     transform: rotate(360deg);
   }
 }
+
 .login {
   display: block;
   background-color: white;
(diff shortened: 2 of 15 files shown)


src/components/Login.js
```diff
@@ -18,59 +20,41 @@ export class Login extends Component {
   handleLogin = event => {
     event.preventDefault();
     const { username, token } = this.state;
+    this.setState({ loading: true})
```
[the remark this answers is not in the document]
hadijahkyampeire: I will look into that but I thought we might use different kinds of loader on different components

mwaz:
LGTM

src/components/Login.js
```diff
@@ -3,11 +3,13 @@ import { notify } from 'react-notify-toast';
 import { connect } from 'react-redux';
 import { Link, withRouter } from 'react-router-dom';
 import { LoginAction } from '../Actions/AuthActions';
+import Loader from './Loader'
 
 export class Login extends Component {
   state = {
     username: '',
     token: '',
+    loading: false
```
kwahalf: lint here, arrange these keys in alphabetical order.

src/components/Navbar.js
```diff
@@ -23,30 +22,18 @@ const Navbar = props => (
       </button>
       <div className="collapse navbar-collapse " id="navbarNav">
         <ul className="navbar-nav ml-auto">
-
-          { props.loggedIn ?
+          {
+            (props.loggedIn === true) && (
```
kwahalf: if props.loggedIn is a boolean, use props.loggedIn&&(......) . instead of props.loggedIn=== true&&(....)

src/styles/index.css
```diff
@@ -0,0 +1,75 @@
...
+    width: 30px;
+    height: 30px;
+    animation: spin 1s linear infinite;
+    display: inline-block;
+  }
+  
+  @keyframes spin {
+    0% { transform: rotate(0deg); }
+    100% { transform: rotate(360deg); }
+  }
```
kwahalf: new line

hadijahkyampeire: Thanks @kwahalf on it

src/components/Navbar.js
hadijahkyampeire: Oh you are right but I did that ===true because it helped solve some error being caused by redux persist but take it us a quick hack that might be cleared after implement some other functionalities

hadijahkyampeire: I have actually resolved the error thank you

src/components/Login.js
hadijahkyampeire: Resolved thanks

hadijahkyampeire:
I have implemented all the feedback, thanks people

src/Actions/AuthActions.js
```diff
@@ -1,27 +1,38 @@
...
+      dispatch(LoginFailed(error.response.data.detail));
+    } else if (error.request) {
+      dispatch(LoginFailed("Request can't be made"));
     }
-  };
+  }
 };
 
-export { LoginAction };
+// logout action
```
dkayiwa: Do we need the above comment?

src/App.css
```diff
@@ -32,63 +32,7 @@
...
-.login {
-  display: block;
-  background-color: white;
-  width: 350px;
-  height: 400px;
-  margin: 10px;
-}
-.inputs {
-  width: 300px !important;
-  margin-left: 20px;
```
dkayiwa: Is removing this file required to fix the login ui?

src/App.css
```diff
@@ -108,28 +52,6 @@
...
 
-.navbar{
-  position: fixed;
-  border-bottom-right-radius: 1em;
-  border-bottom:solid;
-  width:100%;
-  background-color: #495056;
-}
-
-.navbar-brand{
```
dkayiwa: Same question here as for the above. Make sure that any change you make in a pull request is exactly doing what the ticket requires. Otherwise if some of this is simply clean up work, you may need to create a separate ticket and hence pull request, for it.

src/App.css
hadijahkyampeire: It was feedback about putting all the css that affects the index page and dashboard in a new style folder

src/Actions/AuthActions.js
hadijahkyampeire: Resolved

src/App.css
dkayiwa: That is clean up or refactoring. It has nothing to do with fixing the login url in the sense that even if you did not move it around, the login would still be fixed by the other changes. Hence qualifying to require its separate ticket. Not so? Unless moving it in a new style folder is what fixed the login. 😊

src/App.css
hadijahkyampeire: I moved all the css that was affecting the login, and Navbar to a new css file called index.css so as to maintain the best practices. That's why it shows those deletes

src/App.css
hadijahkyampeire: So what should I do because, I already implemented all the feedback from people as fixing the login UI

dkayiwa: Revert those changes that have nothing to do with the actual fix. Then create a separate ticket and pull request for them.

PR #14, opened 2018-05-23
OCLOMRS-27: Adding Nginx Configuration to allow serving of different URL's in the single page application

mwaz:
# JIRA TICKET NAME:
[OCLOMRS-27](https://issues.openmrs.org/browse/OCLOMRS-27)

# Summary:
Currently the Nginx configuration does not allow serving different routes such as **_/dashboard_** or **_/login_**. The default.conf will overwrite the existing Nginx default.conf settings in the docker container and allow Nginx to serve files from all the supplied URL's. The Pull Request is in relation to the Ticket [ITSM-4148](https://issues.openmrs.org/browse/ITSM-4148).


cintiadr:
I'm slightly confused by what exactly you want to achieve. 
So, you are _not_ able to change the the infrastructure nginx proxy. You cannot override or modify anything outside your docker container from inside your docker container, due to technical and logical restrictions. 
Any application-specific logic should be done inside your docker container. I have more than 20 docker containers running now, your docker container should be responsible for any logic required by your application. 

That said, _nginx.conf_ is _not_ a yaml file. It only works for the infrastructure nginx, which happens to be deployed using ansible (which parsers a yaml file). I copied the lines so you could have an idea what our proxy is using. 

What you need to do, instead, is changing the file /etc/nginx/conf.d/default.conf inside your docker container. I don't really understand how you are modifying your docker image without changing your docker file. I did build your docker image locally, and as expected it didn't work. 

First, I'd strongly recommend you update the docker compose file, as I recommended before: https://github.com/openmrs/openmrs-ocl-client/pull/16

And then, I'd ask you to do:

# build docker image
$ docker build -t openmrs/ocl-client:local .

# start docker image
$ docker-compose up -d

# enter docker image
$ docker exec -it openmrs-ocl-client_oclclient_1 sh 

Copy the file  /etc/nginx/conf.d/default.conf to this repo. Edit the way you want. 
Change your dockerfile to copy this file to your container. 

To redo it:
$ docker-compose down -v

# build docker image
$ docker build -t openmrs/ocl-client:local .

# start docker image
$ docker-compose up -d

After you get it working inside your browser, locally, I think it would be good to go.

mwaz:
Thankyou @cintiadr I have made changes to the PR as requested.

cintiadr:
It works! 💃 

When I run it locally, http://localhost:8081/dashboard loads. 

This PR is good as is, but I'd recommend for next changes/PRs:
  - Can you add on the README how to use docker/docker-compose locally? I could do it too, but I have the feeling you'd be able to write it from a beginners' perspective, making the README more... readable.
  - If you want, and only if you want, you could move the default.conf and docker-compose file to a folder; usually it's called like `docker`, for files specifically used for docker. Moving the dockerfile is possible, but it's a little boring. It's just organisation, it doesn't really change functionality.  Note: if you move the docker-compose.yaml file, you need to use -f option in docker-compose command to find the file
  - Again, if you want, you could delete all the comments from inside your default.conf file. It doesn't affect at all.

mwaz:
I have updated the PR with requested changes @cintiadr

emasys:
LGTM

PR #18, opened 2018-05-24
Allow Node tests to be parsed by Bamboo build

cintiadr:
# Summary:
So, I'd like to configure Bamboo to parse logs results from this project, like:
https://pasteboard.co/HmEpmxo.png

Bamboo can parse junit and mocha (mostly), so I'd like to configure junit. 

I do believe it probably won't affect other functionalities.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@
 
 # testing
 /coverage
+junit.xml
 
 # production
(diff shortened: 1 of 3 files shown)


dkayiwa:
Thanks @cintiadr 👍

PR #28, opened 2018-06-11
OCLOMRS-36: paginate search results

emasys:
# JIRA TICKET NAME:
[As a user I want to view all sources with paginations](https://issues.openmrs.org/browse/OCLOMRS-36)

# Summary:
After retrieving all the sources I want to view all sources other than just a few default. This can be handled by paginations


mwaz:
LGTM

PR #42, opened 2018-06-27
OCLOMRS-52: A user should be able to view paginated concepts in a particular source

shakirandagire:
# JIRA TICKET NAME:
[A user should be able to view paginated concepts in a particular source	
](https://issues.openmrs.org/browse/OCLOMRS-52)

# Summary:
As a user, I should be able to view paginated concepts in a particular source

### DIFF
--- a/src/components/dashboard/container/SpecificConcept.jsx
+++ b/src/components/dashboard/container/SpecificConcept.jsx
@@ -1,10 +1,13 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
+import autoBind from 'react-autobind';
+import InfiniteScroll from 'react-infinite-scroll-component';
 import fetchConceptsActionTypes from '../../../redux/actions/concepts/specificConceptAction';
 import '../styles/index.css';
 
 import SideBar from '../components/SideNavigation';
+import { clearConcepts } from '../../../redux/actions/concepts/ConceptActionCreators';
 import SearchConcept from '../components/concepts/Search';
 import SpecificConceptList from '../components/concepts/SpecificConceptList';
 
@@ -15,15 +18,17 @@ export class SpecificConcept extends Component {
       owner: PropTypes.string,
     })).isRequired,
     isFetching: PropTypes.bool.isRequired,
+    clearConcepts: PropTypes.func.isRequired,
+    hasMore: PropTypes.bool.isRequired,
   };
   constructor(props) {
     super(props);
 
     this.state = {
       searchInput: '',
+      offset: 2,
     };
-    this.onSearch = this.onSearch.bind(this);
-    this.onSubmit = this.onSubmit.bind(this);
+    autoBind(this);
   }
 
   componentDidMount() {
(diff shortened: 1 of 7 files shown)


src/components/dashboard/container/SpecificConcept.jsx
```diff
@@ -15,15 +18,17 @@ export class SpecificConcept extends Component {
...
   constructor(props) {
     super(props);
 
     this.state = {
       searchInput: '',
+      offset: 2,
     };
-    this.onSearch = this.onSearch.bind(this);
-    this.onSubmit = this.onSubmit.bind(this);
+    autoBind(this);
```
mwaz: I still think there is a better way of doing autoBind('functionOne', 'functionTwo')

src/tests/Dashboard/reducer/conceptReducer.test.js
```diff
@@ -36,4 +39,29 @@ describe('Test suite for concepts reducer', () => {
       loading: false,
     });
   });
+
+  it('should handle SEARCH_CONCEPTS', () => {
+    expect(reducer(
```
Efosaok: You would want to make your it block messages more descriptive when writing tests for reducers, actions and components. Tests are usually the first place where someone onboarding into a codebase can look into and get an idea of what each unit function is doing, and you'll be helping out if your test case message is the actual description of what that function or unit block of code being tested is actually doing..

a better way you could write this could be

```
it sets redux state key loading and hasMore to false
on dispatching action type SEARCH_CONCEPTS.`
```
overall, I think you have done a great Job.

dkayiwa: I like it so much when the new guys on the team also start reviewing code! 👍

shakirandagire: Thank you for the feedback @Efosaok. I have made the relevant changes to my tests to make them more descriptive as suggested.
@dkayiwa we like this too. Thanks @Efosaok for commenting

src/components/dashboard/container/SpecificConcept.jsx
shakirandagire: @mwaz I chose to use autoBind(this) to help bind the onSearch and onSubmit methods to this instance. I had not navigated the way suggested to do the binding but will explore it in future

mwaz: awesome..

shakirandagire:
@dkayiwa I have fixed the merge conflicts

.travis.yml
```diff
@@ -1,9 +1,11 @@
 language: node_js
 node_js:
   - "8.9.4"
-script:
-  - npm install
+install:
```
dkayiwa: Are these changes required as part of this ticket?

shakirandagire: I had issues with travis build terminating without running tests.
I researched and the solution was to restructure the travis.yml file.

dkayiwa: That requires a separate ticket and pull request. Do you still remember the article i shared with you over the weekend? 😊

PR #45, opened 2018-06-28
OCLOMRS-59: Search through all the concepts in a dictionary

emasys:
# JIRA TICKET NAME:
[As a user, I should be able to search through all the concepts in a dictionary](https://issues.openmrs.org/browse/OCLOMRS-59)

# Summary:
A user should be able to search through all the concepts in a dictionary. Also, a user should be able to filter search results. Filtering should be based on sources and classes.


src/components/dictionaryConcepts/containers/DictionaryConcepts.jsx
```diff
@@ -29,22 +34,110 @@ export class DictionaryConcepts extends Component {
...
   };
 
   constructor(props) {
     super(props);
     this.state = {
       conceptsCount: getTotal(),
+      searchInput: '',
+      filterList: [],
     };
+    autoBind(this);
```
mwaz: maybe developer politics but I still think declaring this to apply to all functions might not be very effective, imagine a scenario where a developer does not know all the functions are bound by this statement. autobind('functionName1', 'functionName2, 'etc') is more recommended.

emasys: You're right. But at the moment all the methods in the `class` in question need to be bound, If that changes in the future then we would consider binding the methods individually. Binding the methods individually now would defeat the purpose of using the `auto-bind` pkg. However, thank you for your feedback. @mwaz

shakirandagire:
LGTM

dkayiwa:
Some merge conflicts here.

src/redux/reducers/ConceptReducers.js
```diff
@@ -1,28 +1,49 @@
-import { IS_FETCHING, FETCH_CONCEPTS, FETCH_DICTIONARY_CONCEPT } from '../actions/types';
-import { filterSources, filterClass } from './util';
+import * as types from '../actions/types';
```
dkayiwa: Did you take a look at this? https://wiki.openmrs.org/display/docs/Java+Conventions#JavaConventions-Donotusewildcardimports

emasys: Fixed. @dkayiwa

PR #47, opened 2018-07-02
OCLOMRS-60: Rework on the side bar

hadijahkyampeire:
# JIRA TICKET NAME:
[OCLOMRS-60: Rework on the side bar](https://issues.openmrs.org/browse/OCLOMRS-60)

# Summary:
Change the sidebar organization.
Indicate the active component.
Change the color of the sidebar.

### DIFF
--- a/src/components/dashboard/components/SideNavigation.jsx
+++ b/src/components/dashboard/components/SideNavigation.jsx
@@ -1,6 +1,7 @@
 import React from 'react';
+import { NavLink } from 'react-router-dom';
 import AddDictionary from './dictionary/AddDictionary';
-import { Link }from 'react-router-dom';
+
 class SideNavigation extends React.Component {
   constructor(props) {
     super(props);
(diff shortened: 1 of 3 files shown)


src/components/dashboard/components/SideNavigation.jsx
```diff
@@ -1,6 +1,14 @@
 import React from 'react';
+import { NavLink } from 'react-router-dom';
 import AddDictionary from './dictionary/AddDictionary';
-import { Link }from 'react-router-dom';
+
+const activeLinkStyles = {
+  backgroundColor: 'gray',
+  borderBottomColor: 'white',
+  color: 'white',
+  borderRadius: '10px',
```
emasys: if you provide a CSS class as a value to the NavLink activeStyle attribute, it would still work. Inline styles are hard to find and edit especially when the codebase eventually becomes really large. I suggest you create a regular CSS class and use it.

PR #49, opened 2018-07-04
OCLOMRS-62: Fix Redux logger logging on production environment

shakirandagire:
# JIRA TICKET NAME:
[OCLOMRS-62: Fix Redux logger logging on production environment](https://issues.openmrs.org/browse/OCLOMRS-62)

# Summary:
The `redux-logger` logs on production environment which should not be so because of security reasons.

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
     "react-tooltip": "^3.6.0",
     "redux": "^4.0.0",
     "redux-logger": "^3.0.6",
-    "redux-mock-store": "^1.5.1",
+    "redux-mock-store": "^1.5.3",
     "redux-test-utils": "^0.2.2",
     "redux-thunk": "^2.2.0"
   },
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -33,12 +33,12 @@
...
     "redux": "^4.0.0",
     "redux-logger": "^3.0.6",
-    "redux-mock-store": "^1.5.1",
+    "redux-mock-store": "^1.5.3",
     "redux-test-utils": "^0.2.2",
     "redux-thunk": "^2.2.0"
   },
   "scripts": {
-    "start": "react-scripts start",
+    "start": "set NODE_ENV=dev && react-scripts start",
```
emasys: I don't know if the new addition to line 41 is necessary, apart from that, PR looks good to me.

shakirandagire: I had put it to explicitly state the environment.
I have since removed it because the environment without it being put in the script

PR #52, opened 2018-07-04
OCLOMRS-64: Implement the UI for Previewing a concept.

hadijahkyampeire:
# JIRA TICKET NAME:
[OCLOMRS-64: Implement the UI for Previewing a concept.](https://issues.openmrs.org/browse/OCLOMRS-64)

# Summary:
- This involves working on the UI that appears when a user selects the concept to be added to a dictionary.
- Create a modal that first shows the user all about the concept they are about to add to their dictionary.

### DIFF
--- a/src/components/dashboard/components/concepts/ConceptModal.js
+++ b/src/components/dashboard/components/concepts/ConceptModal.js
@@ -1,0 +1,41 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const AddConceptModal = props => (
+  <div>
+    <div className="modal fade concept-modal" id={`conceptPreview${props.id}`} tabIndex="-1" role="dialog" aria-labelledby="exampleModalLabel" aria-hidden="true">
+      <div className="modal-dialog" role="document">
+        <div className="modal-content">
+          <div className="modal-header">
+            <h5 className="modal-title" id="exampleModalLabel">{props.display_name} Preview</h5>
+            <button type="button" className="close" data-dismiss="modal" aria-label="Close">
+              <span aria-hidden="true">&times;</span>
+            </button>
+          </div>
+          <div className="modal-body">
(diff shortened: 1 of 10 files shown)


src/components/dashboard/components/concepts/ConceptModal.js
```diff
@@ -0,0 +1,41 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const AddConceptModal = props => (
```
shakirandagire: The modal shows concept details, could you change the name to reflect that instead of AddConceptModal

src/tests/Dashboard/container/ConceptModal.test.js
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+import { mount } from 'enzyme';
+import AddConceptModal from '../../../components/dashboard/components/concepts/ConceptModal';
+
+describe('Concept Modal', () => {
+  it('should render without crushing', () => {
+    const wrapper = mount(<AddConceptModal />);
+    expect(wrapper).toMatchSnapshot();
+  });
```
emasys: You could actually do more assertions here

src/components/dashboard/components/concepts/ConceptModal.js
hadijahkyampeire: It is actually where the user clicks to add that particular concept

src/tests/Dashboard/container/ConceptModal.test.js
hadijahkyampeire: Ideas are welcome but for now, this is enough since its just a component without logic

src/components/dashboard/components/concepts/ConceptModal.js
shakirandagire: Alright then that makes sense

PR #58, opened 2018-07-06
OCLOMRS-78: Move the Add dictionary button to the Dashboard

hadijahkyampeire:
# JIRA TICKET NAME:
[OCLOMRS-78: Move the Add dictionary button to the Dashboard](https://issues.openmrs.org/browse/OCLOMRS-78)

# Summary:
- This involves moving the dictionary button from the side Navigation bar to the Page where dictionaries display.

### DIFF
--- a/src/components/dashboard/container/DictionariesDisplay.jsx
+++ b/src/components/dashboard/container/DictionariesDisplay.jsx
@@ -4,117 +4,132 @@ import propTypes from 'prop-types';
 import InfiniteScroll from 'react-infinite-scroll-component';
 import { connect } from 'react-redux';
 import '../styles/index.css';
-import { fetchDictionaries, searchDictionaries, fetchDictionary } from '../../../redux/actions/dictionaries/dictionaryActionCreators';
+import {
+  fetchDictionaries,
+  searchDictionaries,
+  fetchDictionary,
+} from '../../../redux/actions/dictionaries/dictionaryActionCreators';
 import { clearDictionaries } from '../../../redux/actions/dictionaries/dictionaryActions';
 import SideBar from '../components/SideNavigation';
 import ListDictionaries from '../components/dictionary/ListDictionaries';
 import SearchDictionaries from '../components/dictionary/DictionariesSearch';
 import UserDashboard from '../components/dictionary/user/UserDashboard';
+import AddDictionary from '../components/dictionary/AddDictionary';
 
 export class DictionaryDisplay extends Component {
-    static propTypes = {
-      fetchDictionaries: propTypes.func.isRequired,
-      dictionaries: propTypes.arrayOf(propTypes.shape({
-        dictionaryName: propTypes.string,
-      })).isRequired,
-      isFetching: propTypes.bool.isRequired,
-      searchDictionaries: propTypes.func.isRequired,
-      clearDictionaries: propTypes.func.isRequired,
-      fetchDictionary: propTypes.func.isRequired,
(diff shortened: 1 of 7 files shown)


src/components/dashboard/container/DictionariesDisplay.jsx
```diff
@@ -4,117 +4,132 @@ import propTypes from 'prop-types';
...
+    clearDictionaries: propTypes.func.isRequired,
+    fetchDictionary: propTypes.func.isRequired,
+    organizations: propTypes.arrayOf.isRequired,
+  };
+  constructor(props) {
+    super(props);
+    this.state = {
+      searchInput: '',
+      dictionaryViewOptions: 'user',
+      show: false,
```
IAMOTZ: Hi @hadijahkyampeire 
This state variable naming is somehow vague.
Just by looking at it, I am unable to guess what you are trying to show/hide, but after digging into the codebase I discovered that it is used to toggle the display of the Dictionary Modal.
A variable naming like `showDictionaryModal` would be better, the purpose of the variable is clear from its naming.

src/components/dashboard/container/DictionariesDisplay.jsx
```diff
@@ -4,117 +4,132 @@ import propTypes from 'prop-types';
...
-    onSearch(event) {
-      const { value, name, checked } = event.target;
-      const trueValue = event.target.type === 'checkbox' ? checked : value;
-      this.setState({ [name]: trueValue });
-      if (name === 'searchInput' && value.length <= 0) {
-        this.props.clearDictionaries();
-        this.props.fetchDictionaries(value);
-      }
-    }
+  handleHide = () => this.setState({ show: false });
```
IAMOTZ: What are you trying to hide with this method(`handleHide`)?
Though it is understandable form the content of the function that you are switching the value of a state variable(`show`). It would be better if the name of the method is more explanatory as to what the method is doing, this would ensure clarity of purpose when this method is passed to other components as` props`.

src/components/dashboard/container/DictionariesDisplay.jsx
```diff
@@ -4,117 +4,132 @@ import propTypes from 'prop-types';
...
+  handleShow = () => this.setState({ show: true });
```
IAMOTZ: What are you trying to hide or show with these methods (`handleHide`, `handleShow`)?
Though it is understandable form the content of the methods that you are switching the value of a state variable(`show`). It would be better if the name of the methods is more explanatory as to what the methods are doing, this would ensure clarity of purpose when this methods are passed to other components as` props`.

src/components/dashboard/container/DictionariesDisplay.jsx
hadijahkyampeire: Thanks for that but I think that will be done in another PR for refactoring code since this was for just moving the button and its properties to another page

src/components/dashboard/container/DictionariesDisplay.jsx
hadijahkyampeire: Same for these it will be another PR for refactoring code

IAMOTZ: Understood.

src/components/dashboard/container/DictionariesDisplay.jsx
IAMOTZ: Owk. That's awesome.

PR #59, opened 2018-07-08
OCLOMRS-71: Fix the bug for concepts in a source not displaying when the owner is not an organization

shakirandagire:
# JIRA TICKET NAME:
[OCLOMRS-71: Fix the bug for concepts in a source not displaying when the owner is not an organization](https://issues.openmrs.org/browse/OCLOMRS-71)

# Summary:
At the moment when viewing concepts in a source created by an owner that is not an organization, they are not displayed even when they exist which should not be the case.

### DIFF
--- a/src/components/dashboard/components/SourceCard.jsx
+++ b/src/components/dashboard/components/SourceCard.jsx
@@ -30,7 +30,7 @@ const SourceCard = (source) => {
           </div>
           <div className="description col-12 text-left">
             <p>
-              <a href={`/dashboard/concepts/${owner}/${name}`} className="source-type">
+              <a href={`/dashboard/concepts/${owner_type}/${owner}/${name}`} className="source-type">
                 {active_concepts} concepts,
               </a>{' '}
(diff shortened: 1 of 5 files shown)


src/components/dashboard/components/SourceCard.jsx
```diff
@@ -30,7 +30,7 @@ const SourceCard = (source) => {
           </div>
           <div className="description col-12 text-left">
             <p>
-              <a href={`/dashboard/concepts/${owner}/${name}`} className="source-type">
+              <a href={`/dashboard/concepts/${owner_type}/${owner}/${name}`} className="source-type">
```
dkayiwa: What criteria do you use to tell whether to use owner_type or ownerType?

shakirandagire: @dkayiwa owner_type is the name that was used in the API when labeling the owner type. In the instance above I could not change it.
But where I used ownerType I was following the Javascript conventions for the camel casing.

dkayiwa: Fair enough 😊

PR #62, opened 2018-07-09
OCLOMRS-83: Fix the url to get CIEL concepts.

shakirandagire:
# JIRA TICKET NAME:
[OCLOMRS-83: Fix the url to get CIEL concepts.](https://issues.openmrs.org/browse/OCLOMRS-83)

# Summary:
Fix the url to view CIEL concepts to be added to the dictionary

### DIFF
--- a/src/components/dictionaryConcepts/components/ConceptDropdown.jsx
+++ b/src/components/dictionaryConcepts/components/ConceptDropdown.jsx
@@ -15,7 +15,7 @@ const ConceptDropdown = props => (
         Add existing concepts
       </button>
       <div className="dropdown-menu" aria-labelledby="dropdownMenuButton">
-        <a className="dropdown-item" href="/dashboard/concepts/hadijah315/ICD-10">
+        <a className="dropdown-item" href="/dashboard/concepts/Organization/CIEL/CIEL">
           Add CIEL concepts
         </a>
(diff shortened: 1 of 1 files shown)


src/components/dictionaryConcepts/components/ConceptDropdown.jsx
```diff
@@ -15,7 +15,7 @@ const ConceptDropdown = props => (
         Add existing concepts
       </button>
       <div className="dropdown-menu" aria-labelledby="dropdownMenuButton">
-        <a className="dropdown-item" href="/dashboard/concepts/hadijah315/ICD-10">
+        <a className="dropdown-item" href="/dashboard/concepts/Organization/CIEL/CIEL">
```
emasys: Kindly use react-router's `Link` to avoid full page re-render.

shakirandagire: @emasys I suggest changing it to link needs a different PR because it will affect the tests,  I only had to change the url to a working one.

emasys: I hope we remember to fix this and it doesn't become a `broken window`.

shakirandagire: We shall fix it but thanks for pointing that out

PR #63, opened 2018-07-11
OCLOMRS-77: Add newly created concept to user dictionary

emasys:
# JIRA TICKET NAME:
[As a user, I should see my newly created concept in my dictionary](https://issues.openmrs.org/browse/OCLOMRS-77)

# Summary:
A newly created concept should be automatically added to a user's dictionary from the user's source.

Dev notes:
A newly created concept by default will be added to a source, you are to make sure it is added to collections(dictionary) at the point of creation as well.

### DIFF
--- a/src/components/dictionaryConcepts/containers/CreateConcept.jsx
+++ b/src/components/dictionaryConcepts/containers/CreateConcept.jsx
@@ -43,6 +43,7 @@ export class CreateConcept extends Component {
       names: PropTypes.array,
       descriptions: PropTypes.array,
     }).isRequired,
+    addedConcept: PropTypes.array.isRequired,
   };
   constructor(props) {
     super(props);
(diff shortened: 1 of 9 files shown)


src/components/dictionaryConcepts/containers/CreateConcept.jsx
```diff
@@ -77,8 +78,10 @@ export class CreateConcept extends Component {
         params: { collectionName, type, typeName },
       },
     } = this.props;
-    if (Object.keys(nextProps.newConcept).length) {
-      nextProps.history.push(`/concepts/${type}/${typeName}/${collectionName}`);
+    if (Object.keys(nextProps.newConcept).length && nextProps.addedConcept.length) {
```
EleisonC: Hey, Emma, would you mind if you put "Object.keys(nextProps.newConcept).length" and "nextProps.addedConcept.length"  in separate variables and use them in the if statement. Quite sure would help make the code easy to read.

emasys: Done

PR #66, opened 2018-07-12
OCLOMRS-92: Create a new UI for the navigation bar

EleisonC:
# JIRA TICKET NAME:
[OCLOMRS-92: Create a new UI for the navigation bar](https://issues.openmrs.org/browse/OCLOMRS-92)

# Summary:
A user should be able to navigate to the different components through selection destinations from the navigation bar.
A user should have access to the general search on the navigation bar

### DIFF
--- a/src/components/NavBar/GeneralSearch.jsx
+++ b/src/components/NavBar/GeneralSearch.jsx
@@ -1,0 +1,17 @@
+import React from 'react';
+
+/* eslint-disable */
+const GeneralSearch = ({ onSearch, onSubmit, searchValue }) => {
+  return (
+    <form className="form-inline" id="generalSearchForm">
+      <input 
+      class="form-control mr-sm-2"
+      id="generalSearch"
+      type="search" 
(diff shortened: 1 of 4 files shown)


src/components/NavBar/GeneralSearch.jsx
```diff
@@ -0,0 +1,17 @@
...
+
+/* eslint-disable */
+const GeneralSearch = ({ onSearch, onSubmit, searchValue }) => {
+  return (
+    <form className="form-inline" id="generalSearchForm">
+      <input 
+      class="form-control mr-sm-2"
+      id="generalSearch"
+      type="search" 
+      placeholder="Search"
```
hadijahkyampeire: Since we are limiting it to all dictionaries, I suggest that we make the placeholder more descriptive

EleisonC: That will be taken care of in the next pull request thank you for the feedback

PR #70, opened 2018-07-16
OCLOMRS-58: paginate concepts in a dictionary

emasys:
# JIRA TICKET NAME:
[As a user, I should be able to see all the concepts in my dictionary with proper pagination](https://issues.openmrs.org/browse/OCLOMRS-58)

# Summary:
Return all concepts in a particular dictionary with proper paginations.
each page should not have more than 10 results
each page should have a previous and next button
infinite scroll could be used but not necessarily


src/components/dictionaryConcepts/containers/DictionaryConcepts.jsx
```diff
@@ -37,60 +38,52 @@ export class DictionaryConcepts extends Component {
...
-    }
+    this.setState({
+      collectionName,
+      type,
+      typeName,
+      conceptsCount: nextProps.totalConceptCount,
+    });
   }
 
-  fetchConcepts(query = '', limit = 10, filterParams = null, filterName = null) {
```
dkayiwa: Does the change from limit 10 to 0 mean you are paginating from client instead of server?

emasys: yes, it's faster.

emasys: I also needed to get the total number of concepts at every point in time whether I'm searching or filtering.

dkayiwa: What happens when a server has too many concepts? Do you still want to load all of them to the client?

emasys: yeah, for now, that's the only way to get the total number of concepts while filtering by sources or class.

dkayiwa: Let me ask it this way. What was the 10 originally accomplishing?

emasys: initially, I was fetching just 10 concepts whether a user was searching or filtering.

dkayiwa: That limit parameter was intentionally put in place for that. So ignoring it may result into running out of server memory when the concepts are very many. Not so?

emasys: I think the server can handle it. You may refer to - https://github.com/OpenConceptLab/oclapi/wiki#pagination-and-limiting-the-number-of-returned-results

dkayiwa: Let me put it this way, why do you think they put a provision for the limit parameter?

emasys: I understand what you are saying, my point is I need to get the total number of concepts each time I fetch the concepts from the server, right now it will not be possible if I use the limit parameter.

dkayiwa: So with your change, are you simply getting the total number of concepts? Or are you actually getting the concepts?

emasys: I'm getting the concepts. There's currently no way to get only the total number of concepts when filtering or searching.

dkayiwa: So my point is, the original code was safer because it does not load all the concepts. And this was achieved by the use of the limit parameter. Getting rid of this parameter looks like a design mistake unless you can convince me otherwise.

emasys: The initial code was a quick hack, I only got the total number of concepts in a dictionary because I saved it in the localStorage. The issues with the initial code are as follows: 
- when a filter returns eleven concepts, the user will only see ten with no idea there could be more because there's no way to know if there's actually more.
- The filter sidebar cannot be populated with items that are not available. i.e - if the first ten concepts are from the same source or class, the sidebar will be populated by only that type of source. This will clearly defeat the aim of having the sidebar. 
- There are other reasons why I have to fetch everything first, but I will stop here. Refer to the images below for more clarity about my second point.
with pagination
without pagination

PR #73, opened 2018-07-18
OCLOMRS-100: Remove the side Navigation bar from all components

EleisonC:
# JIRA TICKET NAME:
[OCLOMRS-100: Remove the side Navigation bar from all components](https://issues.openmrs.org/browse/OCLOMRS-100)

# Summary:
In the new navigation bar, links were included to access the different components. They include All dictionaries, sources and concepts. Hence created repetition in the side Navigation bar.
So in this task, the side nav bar will be eliminated from the application to avoid repetition


src/components/dashboard/container/SourceSearch.jsx
```diff
@@ -9,7 +9,7 @@ import { clearSources } from '../../../redux/actions/sources/sourcesActionCreato
 import { fetchingOrganizations } from '../../../redux/actions/dictionaries/dictionaryActionCreators';
 import '../styles/index.css';
 
-import SideBar from '../components/SideNavigation';
+
```
hadijahkyampeire: I think this is too much redundant space, if you don't mind you can remove one line for clean code

hadijahkyampeire: however look good to me

PR #74, opened 2018-07-18
OCLOMRS-84: Redesign user dashboard

emasys:
# JIRA TICKET NAME:
[Redesign the user dashboard to follow the provided mockup](https://issues.openmrs.org/browse/OCLOMRS-84)

# Summary:
As a user, after signing in to the app, I should see a user dashboard similar to what is in the [mockup](https://docs.google.com/document/d/1R_Fgr5SBl4xFNJgj6yMJNVY63b5H_OUqM55o1GFqFKs/edit)


src/components/userDasboard/components/DashboardDetails.jsx
```diff
@@ -0,0 +1,71 @@
...
+    </div>
+  );
+};
+
+DashboardDetails.propTypes = {
+  numberOfOrgs: PropTypes.number.isRequired,
+  numberOfDictionary: PropTypes.number.isRequired,
+  organizations: PropTypes.array.isRequired,
+};
+
```
Efosaok: This component is rendering three different views based on some condition, don't you think it'll be better if these three different views are modularised as it would aid reusability and make this component look more compact.

Each possible view in this component has the **_div tag with className user-data_** **_and a p tag with className lead_**, I think it'll be better if you write this without having to repeat this tags in the different views.

emasys: Thanks for your feedback. However, if you closely look at the component you'd see there's no point modularizing any part of this component. Inasmuch as all the blocks look similar I would like you to take a closer look at the lines below: 
- 12
- 26
- 58

Trying to break this component furthermore will only make it more difficult to understand or edit in the future. 

Thanks again. @Efosaok

Efosaok: sorry I forgot to approve this earlier, I just felt it would be better modularizing this component, but if you insist it's better this way for this project, it's fine as well. as for the templating on those lines, correct me if I am wrong, but I do not see how modularizing it affects that part. the child component could receive the value as props.

PR #76, opened 2018-07-18
OCLOMRS-99: Add more useful information to the login page

hadijahkyampeire:
# JIRA TICKET NAME:
[OCLOMRS-99: Add more useful information to the login page](https://issues.openmrs.org/browse/OCLOMRS-99)

# Summary:
- A user should be able to know how to login to the app.
Add more information about how to log in.
Update the links to use the 'qa' version of the traditional OCL


src/components/login/container/LoginDetails.jsx
```diff
@@ -0,0 +1,30 @@
...
+const LoginDetails = () => (
+  <div className="login-details">
+    <div className="card">
+      <div className="card-header">
+        <strong>What to ensure</strong>
+      </div>
+      <div className="card-body">
+        Ensure you have an account from the traditional OCL. After that,{' '}
+        <a href="https://qa.openconceptlab.org/accounts/login/">
+          Login into the traditional OCL
```
Efosaok: I do not think using the _a_ tag is ideal as it would cause a page to reload which goes against the essence of React, I'd advise you use the Link component provided by react-router-dom.

hadijahkyampeire: ok thanks, will do that

src/components/login/container/LoginDetails.jsx
```diff
@@ -0,0 +1,31 @@
...
+  <div className="login-details">
+    <div className="card">
+      <div className="card-header">
+        <strong>What to ensure</strong>
+      </div>
+      <div className="card-body">
+        Ensure you have an account from the traditional OCL. After that,{' '}
+        <Link to="https://qa.openconceptlab.org/accounts/login/">
+          Login into the traditional OCL
+        </Link>
```
emasys: This should be an `a` tag that opens a new tab so the users won't have to completely leave the app they are trying to learn how to use.

src/components/login/container/LoginDetails.jsx
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+import { Link } from 'react-router-dom';
+import '../styles/index.css';
+import newVideo from '../styles/video/newInstructions.mp4';
+
+const LoginDetails = () => (
+  <div className="login-details">
```
emasys: You might want to wrap the component in a bootstrap `container class` so the page doesn't get distorted on tablets.

src/components/login/container/index.jsx
```diff
@@ -85,13 +85,15 @@ export class Login extends Component {
...
                   <a
-                    href="https://openconceptlab.org/accounts/signup/"
+                    href="https://qa.openconceptlab.org/accounts/signup/"
                     target="_blank"
                     rel="noopener noreferrer"
                   >
                     Sign up
-                  </a>
+                  </a><br />
+                  <a href="/loginDetails" >Having Trouble with Login?</a>
```
emasys: You should use a `Link` here. I believe you know why.

src/components/login/container/LoginDetails.jsx
hadijahkyampeire: I have done that, thanks

src/components/login/container/LoginDetails.jsx
hadijahkyampeire: Done it

src/components/login/container/LoginDetails.jsx
hadijahkyampeire: Hey @emasys also commented on this and actually recommend to change it back to `a` so that's why its back

emasys: This is an external link, you might want to make it open in a new tab

hadijahkyampeire: After thorough research and discussions, we decided to leave the ```a``` for external links, therefore this comment was resolved.

dkayiwa: Does this mean we shall again change from qa.openconceptlab.org back to https://openconceptlab.org?

dkayiwa: What do others think?

hadijahkyampeire: I also agree with @emasys because if we are to change the ```qa``` even the API endpoint needs to change, It looks like a ticket on its own

dkayiwa: Can u create the ticket?

PR #85, opened 2018-07-24
OCLOMRS-109: Fix the links going back to a dictionary after creating and viewing specific concepts

shakirandagire:
# JIRA TICKET NAME:
[OCLOMRS-109: Fix the links going back to a dictionary after creating and viewing specific concepts](https://issues.openmrs.org/browse/OCLOMRS-109)

# Summary:
Currently, after a user creates a custom concept or views specific concepts in the source the links to take you back to the dictionary are throwing a 404(Page not found error).


src/components/dashboard/container/SpecificConcept.jsx
```diff
@@ -101,7 +102,7 @@ export class SpecificConcept extends Component {
         <div className="container-fluid pt-3">
           <div className="row justify-content-center">
             <div className="col-10 offset-sm-1">
-              <Link to={`/concepts/${userType}/${typeName}/${dictionaryId}/${dictionaryName}`} id="dictionary-link">
+              <Link to={`/concepts/${userType}/${typeName}/${dictionaryId}/${dictionaryName}/${language}`} id="dictionary-link">
```
hadijahkyampeire: For us to take care of further changes in the URL, you could just store the pathname because its dynamic. Also, I don't think this solves the whole problem, for example when you view concepts in a source this link will still throw an error.

shakirandagire: Thank you for the feedback.
I have made the relevant changes

PR #86, opened 2018-07-24
OCLOMRS-112: Fix the links from specific concepts going back to a dictionary or sources

hadijahkyampeire:
# JIRA TICKET NAME:
[OCLOMRS-112: Fix the links from specific concepts going back to a dictionary or sources](https://issues.openmrs.org/browse/OCLOMRS-112)

# Summary:
- Currently when a user is viewing concepts within a source, and then tries to go back it throws a 404 and the message displayed is also misleading.

### DIFF
--- a/src/components/dashboard/container/SpecificConcept.jsx
+++ b/src/components/dashboard/container/SpecificConcept.jsx
@@ -1,7 +1,6 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
-import { Link } from 'react-router-dom';
 import autoBind from 'react-autobind';
 import InfiniteScroll from 'react-infinite-scroll-component';
 import fetchConceptsActionTypes from '../../../redux/actions/concepts/specificConceptAction';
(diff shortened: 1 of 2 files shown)


src/components/dashboard/container/SpecificConcept.jsx
```diff
@@ -69,6 +68,16 @@ export class SpecificConcept extends Component {
     }));
   }
 
+  handlePath = (path, dictionarName) => {
```
EleisonC: could you correct the grammatical error on line 71 dictionarName to dictionaryName

src/components/dashboard/container/SpecificConcept.jsx
```diff
@@ -69,6 +68,16 @@ export class SpecificConcept extends Component {
...
+    let pathName;
+    if (path.indexOf('/concepts/') > -1) {
+      pathName = dictionarName;
```
EleisonC: same here Grammatical error

src/components/dashboard/container/SpecificConcept.jsx
hadijahkyampeire: oh thanks

src/tests/Dashboard/container/SpecificConcept.test.jsx
```diff
@@ -134,5 +134,28 @@ describe('Dashboard SpecificConcept Component', () => {
     expect(mapStateToProps(initialState).concepts).toEqual([]);
     expect(mapStateToProps(initialState).isFetching).toEqual(false);
   });
+
+  it('should handle change in path', () => {
```
EleisonC: Do you mind adding the word "the" before path "should handle change in the path"

hadijahkyampeire: Thanks, but I think this is right English because if I put the, It might sound like I am talking of one path, but actually, am testing any change in path. We can stick to this I guess.

PR #88, opened 2018-07-26
OCLOMRS-122: Fix the grammar of the dictionary displayed to the user.

EleisonC:
# JIRA TICKET NAME:
[OCLOMRS-122: Fix the grammar of the dictionary displayed to the user.](https://issues.openmrs.org/browse/OCLOMRS-122)

# Summary:
The home screen currently shows “You currently have 0 personal dictionary” and the title below is “Your dictionary”. Both of these should indicate“dictionaries” (plural).

### DIFF
--- a/src/components/userDasboard/container/UserDashboard.jsx
+++ b/src/components/userDasboard/container/UserDashboard.jsx
@@ -47,7 +47,7 @@ export class UserDashboard extends Component {
       orgDictionary,
       loading,
     } = this.props;
-    const dictionary = public_collections < 2 ? 'dictionary' : 'dictionaries';
+    const dictionary = public_collections === 1 ? 'dictionary' : 'dictionaries';
     return (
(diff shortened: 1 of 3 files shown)


src/components/userDasboard/container/UserDashboard.jsx
```diff
@@ -47,7 +47,7 @@ export class UserDashboard extends Component {
       orgDictionary,
       loading,
     } = this.props;
-    const dictionary = public_collections < 2 ? 'dictionary' : 'dictionaries';
```
emasys: What do you intend to achieve with this change?

EleisonC: Well dictionary will only be shown once when there is only one dictionary

emasys: ok, cool.

PR #89, opened 2018-07-26
OCLOMRS:118: Make the entire dictionary card clickable

emasys:
# JIRA TICKET NAME:
[Make the entire dictionary card clickable](https://issues.openmrs.org/browse/OCLOMRS-118)

# Summary:
Should make the entire card clickable, and it should redirect to the dictionary overview instead of having another option redirecting to the concepts page.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,4 @@ yarn.lock
 #test snapshots
 **/__snapshots__/**
 src/components/dashboard/container/ListContainer.jsx
+.env
--- a/src/App.css
+++ b/src/App.css
@@ -74,3 +74,7 @@
 .error-template {
   margin-top: 10px !important;
 }
+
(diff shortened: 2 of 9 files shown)


.env
```diff
@@ -0,0 +1,2 @@
+REACT_APP_BASE_URL=https://api.qa.openconceptlab.org/
+REACT_APP_TRADITIONAL_OCL=https://qa.openconceptlab.org
```
shakirandagire: Could you add a new line at the end of the file

dkayiwa:
Are all these changes about simply making the card clickable?

emasys:
Yes, @dkayiwa. The card is a reusable component, it was used in two different components. Also, we needed to get rid of other links within it so that only the card and the `view more` button will be clickable.

PR #93, opened 2018-07-27
OCLOMRS-141: Fix a bug on the user dashboard page.

EleisonC:
# JIRA TICKET NAME:
[OCLOMRS-141: Fix a bug on the user dashboard page.](https://issues.openmrs.org/browse/OCLOMRS-141)

# Summary:
Currently, the URL we are using does not allow a user to view all the dictionaries they have created. This is because the URL was designed to only return 10 items and not even adding a limit to the URL would solve the issue.

### DIFF
--- a/src/tests/userDashboard/actions/index.test.js
+++ b/src/tests/userDashboard/actions/index.test.js
@@ -8,15 +8,13 @@ import {
   FETCH_USER_ORGANIZATION,
   GET_USER,
   IS_FETCHING,
-  FETCH_ORG_DICTIONARY,
   CLEAR_DICTIONARY,
 } from '../../../redux/actions/types';
 import {
   fetchUserData,
-  fetchUserDictionary,
+  fetchsUserDictionaries,
   fetchUser,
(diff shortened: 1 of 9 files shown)


src/tests/userDashboard/actions/index.test.js
```diff
@@ -60,7 +58,7 @@ describe('Test suite for user dashboard actions', () => {
     ];
 
     const store = mockStore({});
-    return store.dispatch(fetchUserDictionary('emasys')).then(() => {
+    return store.dispatch(fetchDictionariesUser('chriskala')).then(() => {
```
hadijahkyampeire: The word `fetchDictionariesUser` sounds confusing, is it possible to change it to `fetchUserDictionaries`. And did you need to change it?

EleisonC: @hadijahkyampeire Done

EleisonC: Yes i had to change it

PR #100, opened 2018-07-31
OCLOMRS-125: Fix the row added after the user clicks Add Name/Synonym and Add Description buttons

kmunene:
# JIRA TICKET NAME:
[Fix the row added after the user clicks Add Name/Synonym and Add Description buttons](https://issues.openmrs.org/browse/OCLOMRS-125)

# Summary:
Currently, when “Add name/synonym” or "Add descriptrion" buttons are clicked the row should be added to the bottom instead of the top of the list.


hadijahkyampeire:
@kmunene could you complete the PR name

kmunene:
@hadijahkyampeire Thank you for your feedback. I have done that

PR #104, opened 2018-08-01
OCLOMRS-148: Fix to select one of the supported locales when a dictionary is created in English or Spanish

shakirandagire:
# JIRA TICKET NAME:
[OCLOMRS-148: Fix to select one of the supported locales when a dictionary is created in English or Spanish](https://issues.openmrs.org/browse/OCLOMRS-148)

# Summary:
Currently, when the user creates a dictionary with English or Spanish as the default language,both languages are saved as the supported locales which should not be the case.

### DIFF
--- a/src/components/dashboard/components/dictionary/common/DictionaryModal.jsx
+++ b/src/components/dashboard/components/dictionary/common/DictionaryModal.jsx
@@ -17,6 +17,7 @@ export class DictionaryModal extends React.Component {
         name: '',
         owner: '',
         description: '',
+        default_locale: 'en',
         supported_locales: 'en, es',
         repository_type: 'OpenMRSDictionary',
       },
@@ -68,8 +69,8 @@ export class DictionaryModal extends React.Component {
       errors.public_access =
         'Kindly select the Permissions for your dictionary';
     }
-    if (!data.supported_locales) {
-      errors.supported_locales = 'Kindly select your preferred locale';
+    if (!data.default_locale) {
+      errors.default_locale = 'Kindly select your preferred locale';
     }
     return errors;
   };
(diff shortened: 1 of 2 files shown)


src/components/dashboard/components/dictionary/common/DictionaryModal.jsx
```diff
@@ -17,7 +17,7 @@ export class DictionaryModal extends React.Component {
         name: '',
         owner: '',
         description: '',
-        supported_locales: 'en, es',
+        supported_locales: 'en',
```
hadijahkyampeire: Is this for only `English` and `Spanish` OR it can affect other languages

shakirandagire: It affects only dictionaries created in English anjd Spanish. I have tested out other languages and they are fine

hadijahkyampeire: ok

dkayiwa: Shouldn't the fix have been to select one of the supported locales?

shakirandagire: Let me change the wording of the ticket to reflect this

shakirandagire: @dkayiwa I have made the relevant changes.

dkayiwa: To me, supported locales looks like should take a list of them.

shakirandagire: @dkayiwa Thanks for pointing that out.
The dictionary modal saves the language in the supported locales field. In state, the supported locales default value is "en, es" so whenever a language is saved its value is "en, es". So I changed the supported locales to a single language to fix that.

dkayiwa: Does that mean supported_locales cannot contain more than one locale?

shakirandagire: @dkayiwa It is true, the supported_locales can contain more than one locale.
I have gone through the documentation again and realised that instead of saving the dictionary language in the supported_locales field, we could instead save it in the default_locale field.
This enables us to leave the supported_locale as a list of languages and also have a designated field for the dictionary language as the default_locale.
I have also made the relevant changes to the code to reflect the above mentioned.

PR #107, opened 2018-08-03
OCLOMRS-150: Add functionality to the preview and add buttons

emasys:
# JIRA TICKET NAME:
[Add functionality to the preview and add buttons](https://issues.openmrs.org/browse/OCLOMRS-150)

# Summary:
As a user, when I click on the preview button of a concept, I should see a popup with more information about that concept with a button to add the concept to my dictionary. If I click on any of the add buttons, the corresponding concept should be added to my dictionary.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10397,6 +10397,11 @@
         "prop-types": "^15.6.1"
       }
     },
+    "reactjs-popup": {
+      "version": "1.1.1",
+      "resolved": "https://registry.npmjs.org/reactjs-popup/-/reactjs-popup-1.1.1.tgz",
+      "integrity": "sha512-K07ioLdQkRvASqTCQdue1EwG0uYES5caXuxg2X63vcRXP2t9e6HGhGlC4CqOMlIdrxT/UwAC0nVnBzQ6Yky14g=="
+    },
     "read-pkg": {
       "version": "1.1.0",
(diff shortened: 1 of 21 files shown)


hadijahkyampeire:
@emasys I would like to understand what this PR is about because as if the functionality for the buttons was already implemented

emasys:
Hi @hadijahkyampeire, I tried using the old implementation of the button functionality but I ran into several errors especially when I go directly to the url of the dictionary I was trying to import from. This is because the data stored in the `localStorage` would be different from the new required data or even empty. 
screen shot 2018-08-03 at 2 53 03 pm
So It became necesary to do everything all over again. If there's any other part of the app depending on this - https://github.com/openmrs/openmrs-ocl-client/pull/107/files#diff-e5250e93fd7111f5bc626e539c1f6207 kindly notify me so something will be done about it now.

hadijahkyampeire:
I guess it's only that but we shall see after its merged, one more thing why can't you just modify the implementation within the already existing files @emasys,  wouldn't that be better?

emasys:
@hadijahkyampeire, I tried but starting afresh was a better option for me.

dkayiwa:
@emasys should it be "Descriptions" or "Description"?

emasys:
Hi @dkayiwa, I don't understand your question.

dkayiwa:
I see a label "Descriptions"

emasys:
Oh yeah, thanks for spotting that error, I have corrected it @dkayiwa.

PR #121, opened 2018-08-27
OCLOMRS-19: User should be able to Update/Edit Dictionaries via the user interface

emmabaye:
# JIRA TICKET NAME:
[ User should be able to Update/Edit Dictionaries via the user interface](https://issues.openmrs.org/browse/OCLOMRS-19)

# Summary:
If the users create their dictionaries, they should be able to update or change their information in the dictionaries.
As a user, when I click the edit button on the dictionaries overview page, I should be able to edit the details for I dictionary.


src/components/dashboard/components/dictionary/EditDictionary.jsx
```diff
@@ -0,0 +1,58 @@
...
+import {
+  fetchingOrganizations,
+  editDictionary,
+} from '../../../../redux/actions/dictionaries/dictionaryActionCreators';
+import languages from './common/Languages';
+
+export class EditDictionary extends React.Component {
+  preSelectDefaultLocale = () => {
+    const { dictionary } = this.props;
+    for (let i = 0; i < languages.length; i++) { // eslint-disable-line
```
IAMOTZ: Hi @emmabaye 
Why are you disabling Eslint here?
```
// eslint-disable-line
```

src/components/dashboard/components/dictionary/EditDictionary.jsx
```diff
@@ -0,0 +1,58 @@
...
+      },
+    } = this.props;
+    return this.props.editDictionary(url, data);
+  }
+  render() {
+    return (
+      <DictionaryModal
+        title="Edit Dictionary"
+        buttonname="Update Dictionary"
+        show={this.props.show}
```
IAMOTZ: Looking at this variable naming `show` it does not really communicate the essence of the value the variable is holding. Please improve the variable naming `show` to become something like `showModal` or any other name that properly communicates the purpose of the variable.

src/components/dashboard/components/dictionary/common/DictionaryModal.jsx
```diff
@@ -131,6 +165,9 @@ export class DictionaryModal extends React.Component {
                       placeholder="English"
                       onChange={this.onChange}
                     >
+                      {isEditingDictionary &&
+                        this.props.preSelectDefaultLocale()
+                      }
```
IAMOTZ:
```
{isEditingDictionary &&
   this.props.preSelectDefaultLocale()
}
```
Calling the function `preSelectDefaultLocale` in the render method is not really a good idea, this is becuase the function would be called again every single time this component re-renders.
The function `preSelectDefaultLocale` is looping through a preety large array:
```
preSelectDefaultLocale = () => {
  ...
  for (let i = 0; i < languages.length; i++) { // eslint-disable-line
    ...
  }
  ...
}
```
Calling it multiple times like this is not efficient.
I have two suggestion to improve this piece of functionality:
- Since the result of `preSelectDefaultLocale` is not dependant on any data from this component(`DictionaryModal`), you can just pass the result of `preSelectDefaultLocale` from `EditDictionary.jsx` to this component.
So in `EditDictionary.jsx`, you would have something like this:
```
 <DictionaryModal
   ...
   // before -> preSelectDefaultLocale={this.preSelectDefaultLocale}  
   defaultLocate= {preSelectDefaultLocale()}
   ...
 />
```
- The second suggestion is to keep the function call `preSelectDefaultLocale` outside the render method.

src/components/dashboard/components/dictionary/common/DictionaryModal.jsx
```diff
@@ -131,6 +165,9 @@ export class DictionaryModal extends React.Component {
                       placeholder="English"
                       onChange={this.onChange}
                     >
+                      {isEditingDictionary &&
+                        this.props.preSelectDefaultLocale()
```
IAMOTZ: From the implementation of `preSelectDefaultLocale`:
```
preSelectDefaultLocale = () => {
  const { dictionary } = this.props;
  for (let i = 0; i < languages.length; i++) { // eslint-disable-line
   if (languages[i].value === dictionary.default_locale) {
   return (<option value={languages[i].value} selected>
   {languages[i].label}
   </option>);
  }
 }
 return null;
}
```
There is a possiblity of the function returning a `null` value, what then happens if this function returns `null`? Is the view not going to be affected?

IAMOTZ: The `preSelectDefaultLocale` is only returning one `option` element:
```
preSelectDefaultLocale = () => {
   ...
   if (languages[i].value === dictionary.default_locale) {
     return (<option value={languages[i].value} selected>
       {languages[i].label}
     </option>);
  }
  ...
}
```
Are you sure the user would be able to select a new language with this implementation?

src/components/dashboard/components/dictionary/EditDictionary.jsx
emmabaye: I think it is sufficient for a modal, since hide and show are common characteristics of a modal. I thought the same too but since I am reusing an existing component. I can't rename the variable here and across the application as it does not pertain to the ticket I am working on. If there is any need for that, another ticket would be created.

src/components/dashboard/components/dictionary/common/DictionaryModal.jsx
```diff
@@ -47,6 +70,9 @@ export class DictionaryModal extends React.Component {
     if (Object.keys(errors).length === 0) {
       this.props
         .submit(this.state.data)
+        .then(() => {
+          this.props.modalhide();
+        })
         .catch((error) => {
           if (error.response) {
```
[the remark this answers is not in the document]
emmabaye: Yes, it should. The modal has to remain open if there is an error after submission because it is a form. Any error emanating from the server-side will be displayed via a toast notification. For the sake of user experience, the modal form has to be open so the user can submit the form again.

src/components/dashboard/components/dictionary/common/DictionaryModal.jsx
emmabaye: This implementation  improves  the user experience  by preselecting the option since it is returning existing values for an edit/update form. Therefore the user doesn't have have to change it or any other input element  if that is not what he/she wants to update.

emmabaye: No, it is not going to be affected. Tested and confirmed.

src/components/dashboard/components/dictionary/common/DictionaryModal.jsx
emmabaye: Thanks for the feedback. However, it seems your suggestions doesn't really fit the context of what I achieved using that. That method preselects a dropdown with an item in the array for the sake of improving user experience, this item will be needed before rendering completes. Speaking of efficiency, will you consider Array.map 
 in the render method inefficient? Beside the array doesn't seem large for the speed of a computer.

src/components/dashboard/components/dictionary/EditDictionary.jsx
emmabaye: For flexibility, I need to use the unary operator.

IAMOTZ: _`...For flexibility...`_ 😕 
I believe there is a reason why your team has configured Eslint to not allow unary operators. Considering the fact that you could have used an alternate approach that shouldn't cost you anything and would also make Eslint happy, don't you think this is against your team's convention?

src/components/dashboard/components/dictionary/common/DictionaryModal.jsx
IAMOTZ: Yes, I know that this implementation improves the user experience by preselecting the language option and that is cool 👍.

My question(which i haven't gotten an answer for) was that, **would the user be able to select a new language if he wills to change the language option**?

src/components/dashboard/components/dictionary/common/DictionaryModal.jsx
emmabaye: Thanks again for the feedback. I will look into this.

src/components/dashboard/components/dictionary/EditDictionary.jsx
emmabaye: I've updated the code.

src/components/dashboard/components/dictionary/common/DictionaryModal.jsx
emmabaye: I've updated the code.

src/components/dashboard/components/dictionary/common/DictionaryModal.jsx
emmabaye: I've made the update

src/components/dashboard/components/dictionary/common/DictionaryModal.jsx
IAMOTZ: Nice one 👍.

src/components/dictionaryConcepts/containers/CreateConcept.jsx
```diff
@@ -177,7 +177,8 @@ export class CreateConcept extends Component {
   addDataFromDescription(data) {
     const currentData = this.state.descriptions.filter(description => description.id === data.id);
     if (currentData.length) {
-      const newList = this.state.descriptions.map(description => (description.id === data.id ? data : description));
+      const newList = this.state.descriptions.map(description =>
+        (description.id === data.id ? data : description));
```
alexmochu: Adjust that spacing between the two lines above

alexmochu:
Resolve this conflicts

src/components/dashboard/components/dictionary/common/DictionaryModal.jsx
```diff
@@ -153,8 +192,11 @@ export class DictionaryModal extends React.Component {
                       value={data.public_access}
                     >
                       <option value="" />
-                      <option value="View">Private </option>
+                      {(isEditingDictionary && dictionary.public_access === 'View') &&
+                      <option value="View" selected>Private </option>
+                      }
                       <option value="Edit"> Public </option>
+                      <option value="View">Private </option>
```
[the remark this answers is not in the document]
emmabaye: @alexmochu It has been updated.

src/components/dictionaryConcepts/containers/CreateConcept.jsx
emmabaye: @alexmochu The reason was because it exceeded the maximum length of a line. The update has been made.

PR #125, opened 2018-08-30
OCLOMRS-194: Fix error to enable OCLOMRS-157 failing test pass on CI

emmabaye:
# JIRA TICKET NAME:
[OCLOMRS-194: Fix error to enable OCLOMRS-157 failing test pass on CI](https://issues.openmrs.org/projects/OCLOMRS/issues/OCLOMRS-194)

# Summary:
Fix for failing test for  the build of ticket OCLOMRS-157.


emmabaye:
@muhozi, the reason I deleted package-lock.json was that I was debugging for the error that caused build failure. I thought it might have something to do with version of packages used and also to avoid possible merge conflict later after installing packages again. It can be automatically generated later by running `npm install`.

src/redux/actions/concepts/dictionaryConcepts.js
```diff
@@ -215,18 +215,16 @@ export const updateConcept = (conceptUrl, data, history) => async (dispatch) =>
   try {
```
sbdonald: Is there anywhere you update `isFetching` to false when the call is successful?

emmabaye: `isFetching` is an action creator that is dispatched with an argument of `true` when fetching and an argument of  `false` when response has been received. It is used for indicating API call status for things like AJAX loader icons.

PR #126, opened 2018-08-30
OCLOMRS-179: User should be able to create concepts in organisation dictionary via user interface

emmabaye:
# JIRA TICKET NAME:
[User should be able to create concepts in organisation dictionary via user interface](https://issues.openmrs.org/browse/OCLOMRS-179)

# Summary:
User should be able to create concepts in organisation dictionary they belong to via the user interface


src/components/dictionaryConcepts/containers/DictionaryConcepts.jsx
```diff
@@ -141,18 +143,37 @@ export class DictionaryConcepts extends Component {
...
       },
       location: { pathname },
       concepts,
       filteredClass,
       filteredSources,
       loading,
+      userIsMember,
     } = this.props;
-    const username = typeName === getUsername();
+    const username = typeName === getUsername() || userIsMember;
```
dondrzzy: It looks like this variable takes on a boolean. Const `username` = true. Is there a way you can rename this variable to make it more readable such is `userIsMember`?

src/components/dictionaryConcepts/containers/DictionaryConcepts.jsx
```diff
@@ -141,18 +143,37 @@ export class DictionaryConcepts extends Component {
...
+    const org = {
+      id: (type === 'orgs') ? typeName : '',
```
dondrzzy: `typeName` is being compared to the result of `getUsername` meaning it is a username and of type string. Assigning this to a variable name called `id` seems a bit of a misuse. Could you rename the variable name?

src/redux/actions/user/index.js
```diff
@@ -51,3 +53,19 @@ export const fetchUserData = username => (dispatch) => {
...
+
+export const fetchMemberStatus = url => async (dispatch) => {
+  let response;
+  try {
+    response = await instance.get(url);
+  } catch (error) {
+    if (error.response.status === 403 || error.response.status === 404) {
+      dispatch(isErrored(false, USER_IS_NOT_MEMBER));
+    }
+    return dispatch(isFetching(false));
```
dondrzzy: I think it makes sense to stop the loader then take the success action instead of the reverse unless there is a good reason you must take the success action while the loader still loads. Same goes for the error. Stop the loader then show the error. Instead of showing an error then stop the loader.

src/tests/userDashboard/actions/index.test.js
```diff
@@ -152,4 +155,46 @@ describe('Test suite for user dashboard actions', () => {
       expect(store.getActions()).toEqual(expectedActions);
     });
   });
+  it('should handle USER_IS_MEMBER', () => {
```
dondrzzy: If this is an action then, it should `dispatch` seems more direct. Same goes for the others below.

src/components/dictionaryConcepts/containers/DictionaryConcepts.jsx
emmabaye: Thanks for the feedback. However, I think `id` is the appropriate name because it is unique and also comes as `id` from the data that comes from the server.

src/components/dictionaryConcepts/components/ConceptTable.jsx
```diff
@@ -33,7 +33,9 @@ const ConceptTable = ({ concepts, loading }) => {
               <th scope="col">Actions</th>
             </tr>
           </thead>
-          <tbody>{concepts.map(concept => <TableItem {...concept} key={concept.version} />)}</tbody>
+          <tbody>
+            {concepts.map(concept => <TableItem {...concept} key={concept.version} org={org} />)}
```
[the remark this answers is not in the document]
emmabaye: Sorry, this is outside the scope of the ticket.

src/components/dictionaryConcepts/containers/DictionaryConcepts.jsx
emmabaye: I have made the update.

src/components/dictionaryConcepts/containers/DictionaryConcepts.jsx
emmabaye: I have made the update.

src/redux/actions/user/index.js
emmabaye: I have made the update.

src/components/dictionaryConcepts/components/ConceptTable.jsx
alexmochu: Remove this extra space in <tbody>

emmabaye: Sorry, extra pace in?

src/tests/userDashboard/actions/index.test.js
emmabaye: Thanks for the feedback. I have renamed the description.

PR #127, opened 2018-08-31
OCLOMRS-195: Rework the Search bar UI on the All Dictionaries, Sources and Concepts pages.

alexmochu:
# JIRA TICKET NAME:
[Rework the Search bar UI on the All Dictionaries, Sources and Concepts pages.](https://issues.openmrs.org/browse/OCLOMRS-195)

# Summary:
Currently, search bars UI in the above mentioned pages does not match what is expected in the original mockup from the OCL documentation.
We should redesign them to match the mockup.

### DIFF
--- a/src/components/dashboard/components/SearchBar.jsx
+++ b/src/components/dashboard/components/SearchBar.jsx
@@ -3,115 +3,92 @@ import React from 'react';
 /* eslint-disable */
 const SearchBar = ({ onSearch, onSubmit, searchValue }) => {
   return (
-    <div className="container-fluid">
-      <div className="row justify-content-center">
-        <div className="col-1" />
-        <div className="col-sm-12 col-md-10  mt-2 col-12">
-          <form className="search-bar-wrapper" onSubmit={onSubmit}>
-            <i className="fas fa-search fa-fw mr-1 col-1" />
+    <div className="row justify-content-center search-container">
+      <div className="concept-search-wrapper col-12 col-md-5 col-sm-8">
+        <i className="fas fa-search fa-fw mr-1" />
+        <form id="submit-search-form" onSubmit={onSubmit}>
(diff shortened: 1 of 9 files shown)


src/components/dashboard/components/SearchBar.jsx
```diff
@@ -3,115 +3,92 @@ import React from 'react';
...
-                data-toggle="dropdown"
-                aria-haspopup="true"
-                aria-expanded="false"
-              >
-                Filter by source type
+            <label className="form-check-label" htmlFor="dictionary">
+              Dictionary
+              </label>
+          </div>
+          <div className="form-check ml-3">
```
dondrzzy: You have repeated the same div element `<div className="form-check ml-3">` about 4 times while only changing the attribute values. Would creating a single functional component that returns the same elements and takes in the attributes that vary as props be a cleaner method or does it also result in the same lines of code? Otherwise, this looks good to me.

Michyjones:
LGTM

Michyjones:
LGTM

src/components/dashboard/components/SearchBar.jsx
alexmochu: Using a functional component would have been a great idea, but I decided to re-use the div element in order to avoid breaking the app since this was just an enhancement to how the search bar UI looked before.

src/components/dashboard/components/dictionary/DictionariesSearch.jsx
```diff
@@ -1,47 +1,23 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import Tooltip from 'react-tooltip';
+// import Tooltip from 'react-tooltip';
```
dkayiwa: What are we using the above for?

PR #130, opened 2018-09-04
OCLOMRS-196:Paginate Concepts on the Concepts view page

Michyjones:
# JIRA TICKET NAME:
[OCLOMRS-196: Paginate Concepts on the Concepts view page](https://issues.openmrs.org/browse/OCLOMRS-196)

# Summary:
A user should be able to select the limit of concepts per page.

### DIFF
--- a/src/components/dashboard/container/Concepts.jsx
+++ b/src/components/dashboard/container/Concepts.jsx
@@ -7,6 +7,7 @@ import '../styles/index.css';
 
 import SearchConcept from '../components/concepts/Search';
 import ConceptList from '../components/concepts/ConceptList';
+import Paginations from '../../bulkConcepts/component/Paginations';
 
 export class ConceptSearch extends Component {
   static propTypes = {
@@ -21,6 +22,8 @@ export class ConceptSearch extends Component {
 
     this.state = {
       searchInput: '',
+      currentPage: 1,
+      limit: 50,
     };
     this.onSearch = this.onSearch.bind(this);
     this.onSubmit = this.onSubmit.bind(this);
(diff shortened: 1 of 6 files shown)


Michyjones:
@muhozi  I have fixed the tests.

src/components/dashboard/container/Concepts.jsx
```diff
@@ -7,6 +7,8 @@ import '../styles/index.css';
 
 import SearchConcept from '../components/concepts/Search';
 import ConceptList from '../components/concepts/ConceptList';
+import Paginations from '../../bulkConcepts/component/Paginations';
+
```
alexmochu: This whitespace is irrelevant and should be removed.

Michyjones: Thanks for that observation let me get rid of the whitespace

src/redux/actions/concepts/index.js
```diff
@@ -5,7 +5,7 @@ import { isErrored, fetchConcepts, isFetching } from './ConceptActionCreators';
 const fetchConceptsAction = (
   query = '',
   names = [],
-  limit = 25,
+  limit = 500,
```
emmabaye: May I know why the limit was changed from 25 to 500? Isn't it much?

Michyjones: @emmabaye I wanted to get more concepts so as to test the limit per page because I realized with a limit of 25 I could not tell if it affects the application functionality and also the time taken to fetch the concepts.

kmunene: A suggestion for this would be to provide a user with the ability to change the page limit.

dkayiwa:
@Michyjones do you know why this is still open for 20 days?

Michyjones:
This feature is no longer needed since we will remove the concepts page.

PR #132, opened 2018-09-05
OCLOMRS-153: A user should be able to select “Other Languages” when creating a new dictionary

emmabaye:
# JIRA TICKET NAME:
[A user should be able to select “Other Languages” when creating a new dictionary](https://issues.openmrs.org/browse/OCLOMRS-153)

# Summary:
Currently, user can only select preferred language for default locale. User should be able "other languages" for supported locales.

### DIFF
--- a/src/components/dashboard/components/dictionary/common/DictionaryModal.jsx
+++ b/src/components/dashboard/components/dictionary/common/DictionaryModal.jsx
@@ -1,10 +1,11 @@
 import React from 'react';
 import { connect } from 'react-redux';
 import { Button, Modal, FormGroup, FormControl } from 'react-bootstrap';
+import Select from 'react-select';
 import PropTypes from 'prop-types';
 import { fetchingOrganizations } from '../../../../../redux/actions/dictionaries/dictionaryActionCreators';
 import InlineError from '../messages/InlineError';
-import language from './Languages';
+import languages from './Languages';
 
 export class DictionaryModal extends React.Component {
   constructor(props) {
@@ -22,6 +23,7 @@ export class DictionaryModal extends React.Component {
         repository_type: 'OpenMRSDictionary',
       },
       errors: {},
+      supportedLocalesOptions: [],
     };
   }
   componentDidMount() {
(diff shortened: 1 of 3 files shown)


src/components/dashboard/components/dictionary/common/DictionaryModal.jsx
```diff
@@ -43,6 +46,19 @@ export class DictionaryModal extends React.Component {
       },
     });
   };
+
+  onChangeSupportedLocale = (arr) => {
```
dondrzzy: Hey @emmabaye , please name this function name the right way. Like a `do action` such as `changeSupportLocale`. `onChangeSupportLocale` is preferable for an event.

src/components/dashboard/components/dictionary/common/DictionaryModal.jsx
```diff
@@ -82,9 +99,30 @@ export class DictionaryModal extends React.Component {
...
+   const {
+     dictionary: {
+       supported_locales,
+     },
+   } = this.props;
+   if (supported_locales === undefined) {
+     return null;
+   }
+   const supportedLocalesOptions = [];
+   for (let i = 0; i < supported_locales.length; i += 1) {
```
dondrzzy: You could have used the short form i++ but that is also okay.

emmabaye: Thanks for the feedback. However eslint flags it if I use i++.

src/components/dashboard/components/dictionary/common/DictionaryModal.jsx
emmabaye: Thanks for the feedback. However,  that is the function that is called when an `option` in the `select` component changes. It is called when an `onChange` event occurs as you can see [here on line 231](https://github.com/openmrs/openmrs-ocl-client/blob/42d13b140e5cdac842ff87de97f244f200392f30/src/components/dashboard/components/dictionary/common/DictionaryModal.jsx).

emmabaye: I have made the changes.

Michyjones:
LGTM

PR #145, opened 2018-09-14
OCLOMRS-221 : Rework the Back to page links to be consistent across the app

alexmochu:
# JIRA TICKET NAME:
[Rework the Back to page links to be consistent across the app](https://issues.openmrs.org/browse/OCLOMRS-221)

# Summary:
Currently, the Back to page links UI is not consistent across the app.
Therefore the Back to page links UI should be made consistent across the app.

### DIFF
--- a/src/components/dashboard/components/dictionary/DictionaryDetailCard.jsx
+++ b/src/components/dashboard/components/dictionary/DictionaryDetailCard.jsx
@@ -47,6 +47,17 @@ const DictionaryDetailCard = (props) => {
 
   return (
     <div className="container">
+      <section className="backLink">
+        <Link
+          to="/dashboard"
+          className="backLinkText"
+        >
+          <span align="left">
+            <i className="fas fa-arrow-left" />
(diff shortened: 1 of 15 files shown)


src/components/dashboard/components/dictionary/DictionaryDetailCard.jsx
```diff
@@ -114,23 +125,6 @@ const DictionaryDetailCard = (props) => {
               Go to concepts
             </Link>
           </fieldset>
-          <fieldset>
```
kmunene: Why was this deleted? Was it part of the issue scope?

alexmochu: @kmunene Thank you for making that observation. No, it was not part of this issue scope. As agreed by the team, I have removed those changes that were fixing a couple of bugs and I have created a ticket to specifically handle that here: https://issues.openmrs.org/browse/OCLOMRS-223

PR #150, opened 2018-09-21
OCLOMRS-199: Search bar and the checkbox while adding CIEL concepts should work asynchronously

emmabaye:
# JIRA TICKET NAME:
[Search bar and the checkbox while adding CIEL concepts should work asynchronously](https://issues.openmrs.org/browse/OCLOMRS-199)

# Summary:
Results shown in table should correspond to the items selected with the checkboxes and the search input value.

### DIFF
--- a/src/components/bulkConcepts/container/BulkConceptsPage.jsx
+++ b/src/components/bulkConcepts/container/BulkConceptsPage.jsx
@@ -61,10 +61,11 @@ export class BulkConceptsPage extends Component {
       target: { name },
     } = event;
     const targetName = name.split(',');
+    const query = `q=${this.state.searchInput}`;
     if (targetName[1] === 'datatype') {
-      this.props.addToFilterList(targetName[0], 'datatype');
+      this.props.addToFilterList(targetName[0], 'datatype', query);
     } else {
-      this.props.addToFilterList(targetName[0], 'classes');
+      this.props.addToFilterList(targetName[0], 'classes', query);
     }
   };
 
--- a/src/components/dictionaryConcepts/containers/DictionaryConcepts.jsx
+++ b/src/components/dictionaryConcepts/containers/DictionaryConcepts.jsx
@@ -95,20 +95,23 @@ export class DictionaryConcepts extends Component {
     const eventAction = event.target.type === 'checkbox' ? checked : value;
     const inputName = name.split(',')[0];
     const filterType = name.split(',')[1];
+    const { typeName, collectionName, searchInput } = this.state;
     if (type === 'checkbox' && filterType === 'source') {
       this.props.filterBySource(
         inputName,
         this.state.type,
-        this.state.typeName,
-        this.state.collectionName,
+        typeName,
+        collectionName,
+        searchInput,
       );
     }
     if (type === 'checkbox' && filterType === 'classes') {
       this.props.filterByClass(
         `"${inputName}"`,
         this.state.type,
-        this.state.typeName,
-        this.state.collectionName,
+        typeName,
+        collectionName,
+        searchInput,
       );
     }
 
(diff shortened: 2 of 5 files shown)


src/components/bulkConcepts/container/BulkConceptsPage.jsx
```diff
@@ -61,10 +61,11 @@ export class BulkConceptsPage extends Component {
       target: { name },
     } = event;
     const targetName = name.split(',');
+    const query = `q=${this.state.searchInput}`;
     if (targetName[1] === 'datatype') {
-      this.props.addToFilterList(targetName[0], 'datatype');
+      this.props.addToFilterList(targetName[0], 'datatype', query);
```
Rhotimee: It'll be nice to destructure addToFilterList since you are using in more than one place.

src/components/bulkConcepts/container/BulkConceptsPage.jsx
```diff
@@ -61,10 +61,11 @@ export class BulkConceptsPage extends Component {
...
     } else {
-      this.props.addToFilterList(targetName[0], 'classes');
+      this.props.addToFilterList(targetName[0], 'classes', query);
```
Rhotimee: destructure addToFilterList

src/components/dictionaryConcepts/containers/DictionaryConcepts.jsx
```diff
@@ -101,6 +101,7 @@ export class DictionaryConcepts extends Component {
         this.state.type,
         this.state.typeName,
         this.state.collectionName,
+        this.state.searchInput,
```
Rhotimee: It'll be cool to destructure this.state here. 
like
`const { type, typeName, collectionName, searchInput } = this.state`

kmunene: I would also advise the same.

src/components/dictionaryConcepts/containers/DictionaryConcepts.jsx
```diff
@@ -109,6 +110,7 @@ export class DictionaryConcepts extends Component {
...
```
kmunene: Same to this. Try to use destructuring.

src/setupTests.js
```diff
@@ -12,3 +12,5 @@ const localStorageMock = {
 const uuid = { v4: jest.fn(() => 1) };
 global.uuid = uuid;
 global.localStorage = localStorageMock;
+
+jasmine.DEFAULT_TIMEOUT_INTERVAL = 30000;
```
kmunene: Why are we setting a timeout here?

src/components/dictionaryConcepts/containers/DictionaryConcepts.jsx
emmabaye: Thanks for the feedback, @kmunene . I have made the update.

src/setupTests.js
emmabaye: The reason for the timeout is to give enough time for the async `login test` to pass instead of pushing to travis over again.

kmunene: Will it affect timeouts set?

src/components/bulkConcepts/container/BulkConceptsPage.jsx
emmabaye: Thanks for the feedback. However, it can't be used here as `addToFilterList` has already been declared in the upper scope.

src/components/bulkConcepts/container/BulkConceptsPage.jsx
Rhotimee: Alright then.

src/components/bulkConcepts/container/BulkConceptsPage.jsx
Rhotimee: Alright then.

src/setupTests.js
emmabaye: No, it won't it specific for the tests.

PR #151, opened 2018-09-26
OCLOMRS-230:Change rename dashboard to Home and the route

kmunene:
# JIRA TICKET NAME:
[Change rename dashboard to Home and the route](https://issues.openmrs.org/browse/OCLOMRS-230)

# Summary:
On the navbar,  the name Dashboard should be replaced with Home. The routes should be also descriptive.

### DIFF
--- a/src/components/Auth/index.jsx
+++ b/src/components/Auth/index.jsx
@@ -23,12 +23,13 @@ export const AuthWrapper = (WrappedComponent) => {
     };
 
     componentWillMount() {
+      const { push } = this.props.history;
       if (!this.props.loggedIn) {
-        this.props.history.push('/');
+        push('/');
         return;
       }
       if (this.props.location.pathname === '/') {
-        this.props.history.push('/dashboard');
+        push('/home');
       }
     }
 
--- a/src/components/Navbar.js
+++ b/src/components/Navbar.js
@@ -39,10 +39,10 @@ export class Navbar extends Component {
                 <li className="nav-item nav-link" >
                   <a
                     className="nav-link text-white"
-                    href="/dashboard"
+                    href="/home"
                   >
                     <strong>
-                      Dashboard
+                      Home
                     </strong>
                   </a>
                 </li>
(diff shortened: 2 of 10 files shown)


src/components/Auth/index.jsx
```diff
@@ -28,7 +28,7 @@ export const AuthWrapper = (WrappedComponent) => {
         return;
       }
       if (this.props.location.pathname === '/') {
-        this.props.history.push('/dashboard');
+        this.props.history.push('/home');
```
dondrzzy: @kmunene, please destructure.

src/components/Navbar.js
```diff
@@ -39,10 +39,10 @@ export class Navbar extends Component {
                 <li className="nav-item nav-link" >
                   <a
                     className="nav-link text-white"
-                    href="/dashboard"
+                    href="/home"
```
dondrzzy: Shouldn't this be a `NavLink`?

src/components/Auth/index.jsx
kmunene: Ok will do that.

src/components/Navbar.js
dondrzzy: I don't think CSS changes the anchor tag's purpose at all but this looks like it's out of your scope.

src/components/login/container/index.jsx
```diff
@@ -30,7 +30,7 @@ export class Login extends Component {
 
     if (loggedIn) {
       notify.show(`Logged in successfully as ${this.state.username}`, 'success', 3000);
-      this.props.history.push('/dashboard');
+      this.props.history.push('/home');
```
dondrzzy: Please use destructuring props assignment.

kmunene: Thanks for the observation. Will do that.

src/components/Navbar.js
muhozi: @kmunene I think that It would be better if you change this anchor tag to `Link` (a `react router` component) because the link created by this anchor will cause the full page to reload if it is clicked. This is out of the scope of this PR but I think that it is a simple bug(bad practice) that needs to be fixed.

kmunene: I think creating a separate task for this would be the right thing to do since the task description has nothing to do with the Navigation bar. This can be done together with a task say make the navigation bar look better. What do you think?

PR #161, opened 2018-10-04
OCLOMRS-241 : Add placeholder on the select multiple option

Duncanian:
# JIRA TICKET NAME:
[Add placeholder on the select multiple option](https://issues.openmrs.org/browse/OCLOMRS-241)

# Summary:
Select multiple option fields should have a placeholder to guide the user on an option to choose.

### DIFF
--- a/src/components/dashboard/components/dictionary/common/DictionaryModal.jsx
+++ b/src/components/dashboard/components/dictionary/common/DictionaryModal.jsx
@@ -184,11 +184,9 @@ export class DictionaryModal extends React.Component {
                       type="text"
                       id="preferred_source"
                       name="preffered_source"
-                      placeholder="CIEL"
                       onChange={this.onChange}
                       value={data.preffered_sources}
                     >
-                      <option value="" />
                       <option value="CIEL">CIEL (default source)</option>
                       <option value="PIH">PIH</option>
                       {
@@ -253,7 +251,6 @@ export class DictionaryModal extends React.Component {
                       onChange={this.onChange}
                       value={data.public_access}
                     >
-                      <option value="" />
                       {(isEditingDictionary && dictionary.public_access === 'View') &&
                       <option value="View" selected>Private </option>
                       }
@@ -324,6 +321,7 @@ export class DictionaryModal extends React.Component {
                       componentClass="textarea"
                       id="dictionary_description"
                       name="description"
+                      placeholder="e.g Description of this dictionary"
                       onChange={this.onChange}
                       value={data.description}
                     />


src/components/dashboard/components/dictionary/common/DictionaryModal.jsx
```diff
@@ -324,6 +321,7 @@ export class DictionaryModal extends React.Component {
                       componentClass="textarea"
                       id="dictionary_description"
                       name="description"
+                      placeholder="e.g This is my organisation's dictionary"
```
emmabaye: Good job. However, I think a better phrase can be used for the placeholder that tells the user the use of this field.

Duncanian: Perhaps you can give me an example from the way you would like it to be.

emmabaye: `Description of this dictionary`, because a dictionary can belong to a user or an organisation.

src/components/dashboard/components/dictionary/common/DictionaryModal.jsx
```diff
@@ -184,11 +184,9 @@ export class DictionaryModal extends React.Component {
                       type="text"
                       id="preferred_source"
                       name="preffered_source"
-                      placeholder="CIEL"
                       onChange={this.onChange}
                       value={data.preffered_sources}
                     >
-                      <option value="" />
```
dkayiwa: Has this change anything to do with the ticket?

src/components/dashboard/components/dictionary/common/DictionaryModal.jsx
```diff
@@ -253,7 +251,6 @@ export class DictionaryModal extends React.Component {
                       onChange={this.onChange}
                       value={data.public_access}
                     >
-                      <option value="" />
```
dkayiwa: Has this change anything to do with the ticket?

src/components/dashboard/components/dictionary/common/DictionaryModal.jsx
Duncanian: Yeah, it has changed as it was initially empty so I removed it to give the preferred sources a placeholder

src/components/dashboard/components/dictionary/common/DictionaryModal.jsx
Duncanian: Yeah, it has changed as it was initially empty so I removed it to give the public access a placeholder.

PR #171, opened 2018-10-12
Dictionary cards should be the same across the application

muhozi:
# JIRA TICKET NAME:
[Dictionary cards should be the same across the application](https://issues.openmrs.org/browse/OCLOMRS-249)

# Summary:
The dictionary cards don't look the same for both homepage and dictionaries page, this has to be fixed. Regarding dictionary cards, divider line on home page that is in dictionaries section has to be aligned with the dictionary cards.

### DIFF
--- a/src/components/userDasboard/container/UserDashboard.jsx
+++ b/src/components/userDasboard/container/UserDashboard.jsx
@@ -57,11 +57,11 @@ export class UserDashboard extends Component {
     } = this.props;
     const dictionary = public_collections === 1 ? 'dictionary' : 'dictionaries';
     return (
-      <div className="container-fluid mt-5">
+      <div className="container">
         <Title title="Home" />
         <AddDictionary show={this.state.show} handleHide={this.handleHide} />
         <div className="row justify-content-center">
-          <div className="col-11 user-info">
+          <div className="col-12 user-info">
             <div className="row">
               <div className="greetings">
                 <h5>Welcome {name}</h5>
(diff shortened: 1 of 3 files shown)


muhozi:
@Rhotimee Yes, it is used.

Rhotimee:
> Hey @Rhotimee Yes, it is used.

p-1, p-2, p-3 are bootstrap classes. I don't think it a good practice to style a bootstrap class. 
I think it'll be better if you use custom classname and style it with CSS here.

muhozi:
> > Hey @Rhotimee Yes, it is used.
> 
> p-1, p-2, p-3 are bootstrap classes. I don't think it a good practice to style a bootstrap class.
> I think it'll be better if you use custom classname and style it with CSS here.

@Rhotimee To remove them or change those class names was not the scope of the ticket. Yeah, you're right. It would be better to use the custom class names, I think that we should create a ticket for this.

Rhotimee:
> > > Hey @Rhotimee Yes, it is used.
> > 
> > 
> > p-1, p-2, p-3 are bootstrap classes. I don't think it a good practice to style a bootstrap class.
> > I think it'll be better if you use custom classname and style it with CSS here.
> 
> @Rhotimee To remove them or change those class names was not the scope of the ticket. Yeah, you're right. It would be better to use the custom class names, I think that we should create a ticket for this.

Ok. Cool. It'll be nice if this can be addressed in another ticket. 

LGTM

src/components/userDasboard/container/UserDashboard.jsx
```diff
@@ -43,7 +43,7 @@ export class UserDashboard extends Component {
 
   gotoDictionary = (url) => {
     this.props.history.push(url);
-  }
```
dondrzzy: @muhozi, Is this a formatting change?

muhozi: @dondrzzy Thanks for pointing this out. This is not related to the ticket. I reverted this change.

PR #176, opened 2018-10-15
OCLOMRS-259 : Copy concepts from another dictionary action should be removed from the dictionary overview

alexmochu:
# JIRA TICKET NAME:
[Copy concepts from another dictionary action should be removed from the dictionary overview](https://issues.openmrs.org/browse/OCLOMRS-259)

# Summary:
Copy concepts from another dictionary action should be removed from dictionary overview page since it's being implemented when creating a new dictionary. [HERE](https://github.com/openmrs/openmrs-ocl-client/pull/146)

### DIFF
--- a/src/components/dashboard/components/dictionary/DictionaryDetailCard.jsx
+++ b/src/components/dashboard/components/dictionary/DictionaryDetailCard.jsx
@@ -147,7 +147,6 @@ const DictionaryDetailCard = (props) => {
             <legend>Actions</legend>
             <ul>
               <li><a href={traditionalUrl} target="_blank" rel="noopener noreferrer"><i className="fas fa-external-link-alt" />Browse in traditional OCL</a></li>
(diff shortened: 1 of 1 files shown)


dondrzzy:
@alexmochu, Is it a must for someone to copy the concepts while creating the dictionary?

alexmochu:
@dondrzzy That is a very good observation. No it is not a must to copy concepts from a dictionary when creating a new dictionary. This is just an option. If you want to copy concepts after you have already created a dictionary you can use the "Add existing concepts" option. The copy concepts from a dictionary action is not necessary on the dictionary overview page.

emmabaye:
> @alexmochu, Is it a must for someone to copy the concepts while creating the dictionary or could someone do that at a later point in time? In which case, this would still be required.

It is not a must. It can be done at a later point in time.

PR #183, opened 2018-10-19
OCLOMRS-261: Change Logout from a link to a button

abulojoshua1:
# JIRA TICKET NAME:
[Change Logout from a link to a button](https://issues.openmrs.org/browse/OCLOMRS-261)

# Summary:
On the navigation bar, log out on the dropdown is a link. This is not correct and as it doesn't lead to another page. This should, therefore, be changed to a button.

### DIFF
--- a/src/components/Navbar.js
+++ b/src/components/Navbar.js
@@ -64,9 +64,11 @@ export class Navbar extends Component {
                     {''} {localStorage.getItem('username') || this.props.user.username} {''}{' '}
                   </a>
                   <div className="dropdown-menu" aria-labelledby="navbarDropdown" id="navbarDropdown">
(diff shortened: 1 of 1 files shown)


src/components/Navbar.js
```diff
@@ -72,11 +72,11 @@ export class Navbar extends Component {
                     </strong>
                   </a>
                   <div className="dropdown-menu" aria-labelledby="navbarDropdown" id="navbarDropdown">
-                    <Link className="dropdown-item nav-link" to="!#" onClick={this.logoutUser}>
+                    <button className="dropdown-item nav-link" to="!#" onClick={this.logoutUser}>
```
dondrzzy: Remove the `to` attribute.

PR #190, opened 2018-10-23
OCLOMRS-280: Change dictionary visibility options to Private and Public

abulojoshua1:
# JIRA TICKET NAME:
[Change dictionary visibility options to Private and Public](https://issues.openmrs.org/browse/OCLOMRS-280)

# Summary:
In create new dictionary form, visibility options should be Private or public instead of View and None

The options in the UI should say “Public” and “Private”
In the back end API these map to View and None, but the end-user of our app shouldn’t ever see this
Also, there should not be a blank option for this. It’s a required field. (User would have it default to Public)

### DIFF
--- a/src/components/dashboard/components/dictionary/common/DictionaryModal.jsx
+++ b/src/components/dashboard/components/dictionary/common/DictionaryModal.jsx
@@ -15,7 +15,7 @@ export class DictionaryModal extends React.Component {
       data: {
         id: '',
         preferred_source: 'CIEL',
-        public_access: '',
+        public_access: 'View',
         name: '',
         owner: '',
         description: '',
@@ -156,10 +156,6 @@ export class DictionaryModal extends React.Component {
     if (!data.owner) {
       errors.owner = 'Kindly select the Owner of the Dictionary';
     }
-    if (!data.public_access) {
-      errors.public_access =
-        'Kindly select the Permissions for your dictionary';
-    }
     if (!data.default_locale) {
       errors.default_locale = 'Kindly select your preferred locale';
     }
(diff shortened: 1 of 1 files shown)


Michyjones:
Fix the failing build

abulojoshua1:
It is the coverage drop that I have been trying to fix as It has dropped by -0.01

abulojoshua1:
I am working on the coverage drop now.

src/components/dashboard/components/dictionary/common/DictionaryModal.jsx
```diff
@@ -156,10 +156,6 @@ export class DictionaryModal extends React.Component {
     if (!data.owner) {
       errors.owner = 'Kindly select the Owner of the Dictionary';
     }
-    if (!data.public_access) {
```
dkayiwa: Could you explain how this change is related to simply changing from View, None, to Public, Private?

src/components/dashboard/components/dictionary/common/DictionaryModal.jsx
```diff
@@ -270,12 +266,8 @@ export class DictionaryModal extends React.Component {
                       onChange={this.onChange}
                       value={data.public_access}
                     >
-                      <option value="" />
-                      {(isEditingDictionary && dictionary.public_access === 'None') &&
```
dkayiwa: Can you explain why this logic was taken away instead of simply changing the option label from None to Private?

src/components/dashboard/components/dictionary/common/DictionaryModal.jsx
abulojoshua1: By giving the Visibility select dropdown a default value, It is impossible for it to be left blank making that verification code redundant. So I did away with it. Previously, the options were only placed in a local state with an OnChange listener meaning that one had to manually make a selection for them to be added in state.

src/components/dashboard/components/dictionary/common/DictionaryModal.jsx
abulojoshua1: Kindly let me look into this issue before I can give you my final conclusion.

abulojoshua1: Please correct me in case I am wrong, but after my review, I do not think that this block of code is necessary due to the fact that the default value on the select element is `public`. One can still choose one of the options when editing a dictionary.

PR #199, opened 2018-10-30
OCLOMRS-271: Fix bad UX issues while escaping the create new dictionary form modal

muhozi:
# JIRA TICKET NAME:
[Fix bad UX issues while escaping the create new dictionary form modal](https://issues.openmrs.org/browse/OCLOMRS-271)

# Summary:
Bad UX: Create a new dictionary, and add another language. The multi-select widget remains open, so I press Esc (thinking this will close the multi-select) but this closes the entire modal popup and discards my info. Even clicking outside of the modal in the gray area closes it and discards the form. This is bad UX for a form in a modal. Instead, the modal should only be dismissable via the Cancel button at the bottom of the form.

In fact, the form is not entirely cleared when closing the form (either via Cancel or clicking outside the modal): the Visibility and Dictionary Name fields stay filled, but the Other Languages multi-select is cleared. If the form is canceled, all data should be discarded.

### DIFF
--- a/src/components/dashboard/components/dictionary/common/DictionaryModal.jsx
+++ b/src/components/dashboard/components/dictionary/common/DictionaryModal.jsx
@@ -56,9 +56,7 @@ export class DictionaryModal extends React.Component {
     if (Object.keys(errors).length === 0) {
       this.props
         .submit(this.state.data)
-        .then(() => {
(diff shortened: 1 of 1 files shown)


Tittoh:
LGTM

PR #213, opened 2018-11-08
OCLOMRS-294: Clean up application dependencies

abulojoshua1:
# JIRA TICKET NAME:
[Clean up application dependencies](https://issues.openmrs.org/browse/OCLOMRS-294)

# Summary:
Currently, in the application dependencies file(`package.json`), app development dependencies are mixed with app dependencies and there are some unused dependencies listed there. This should be worked on and do and clean up.

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
     "eslint-plugin-import": "^2.14.0",
     "eslint-plugin-jsx-a11y": "^6.1.2",
     "eslint-plugin-react": "^7.11.1",
-    "jest-junit": "^5.2.0",
+    "jest-junit": "^4.0.0",
     "moxios": "^0.4.0",
     "node-sass-chokidar": "^1.3.3",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -46,7 +46,7 @@
     "eslint-plugin-import": "^2.14.0",
     "eslint-plugin-jsx-a11y": "^6.1.2",
     "eslint-plugin-react": "^7.11.1",
-    "jest-junit": "^5.2.0",
+    "jest-junit": "^4.0.0",
```
Duncanian: What's the reason for the downgrade?

abulojoshua1: The newer version causes an error saying I should configure a reporter in jest when I run `npm run test:bamboo`. But configuring a new reporter only conflicts with what jest uses when the app was created using Create React App. So I had to downgrade to the one that was previously there

Duncanian: Awesome! I get it now.

PR #220, opened 2018-11-14
OCLOMRS-315: Make the subscription modal background to be lighter

abulojoshua1:
# JIRA TICKET NAME:
[Make the subscription modal background to be lighter](https://issues.openmrs.org/browse/OCLOMRS-315)

# Summary:
The modal background on the subscription modal is too dark because multiple modals get stacked on top of each other depending on the number of `subscription URL` links. This gives the modal background an effect that makes it appear to be dark. Only a single modal should be displayed.

### DIFF
--- a/src/components/dashboard/components/dictionary/DictionaryDetailCard.jsx
+++ b/src/components/dashboard/components/dictionary/DictionaryDetailCard.jsx
@@ -3,6 +3,7 @@ import PropTypes from 'prop-types';
 import { Link } from 'react-router-dom';
 import DictionaryVersionsTable from './DictionaryVersionsTable';
 import ReleaseVersionModal from './common/ReleaseVersionModal';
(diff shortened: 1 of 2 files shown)


src/styles/app.scss
```diff
@@ -125,3 +125,7 @@
 #release-head {
   font-family: 'Roboto', sans-serif;
 }
+
+.modal-backdrop {
+  background-color: #7e8081 !important;
+}
```
muhozi: @abulojoshua1 Why are you using bootstrap class to change this. Don't you think that this can affect other parts of the app where this class was used?

abulojoshua1: @muhozi, correct me if I am wrong, my goal of this task was to not only make the `subscription modal` background lighter but to also make all modal background colors consistent.

PR #221, opened 2018-11-14
OCLOMRS-314: Fix action buttons on add bulk concepts page

muhozi:
# JIRA TICKET NAME:
[Fix action buttons on add bulk concepts page](https://issues.openmrs.org/browse/OCLOMRS-314)

# Summary:
On add bulk concepts page, the bottom action buttons(**Add all selected** and **Back** buttons) don't appear properly.

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { Fragment } from 'react';
 import { Provider } from 'react-redux';
 import { BrowserRouter, Route, Switch } from 'react-router-dom';
 import Navbar from './components/Navbar';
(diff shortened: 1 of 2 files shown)


src/App.js
```diff
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { Fragment } from 'react';
```
Duncanian: Why is this preferred over the normal `div`?

muhozi: @Duncanian I preferred `Fragment` because it doesn't add extra nodes to the DOM  but `div` does.

PR #223, opened 2018-11-14
OCLOMRS-317: Make the released versions table actions links look more  distinct from each other

Michyjones:
# JIRA TICKET NAME:
[OCLOMRS-317: Make the released versions table actions links look more  distinct from each other](https://issues.openmrs.org/browse/OCLOMRS-317)

# Summary:
There are three links in released versions table, but they appear as one link. After this the link should be distinctive.


Michyjones:
@muhozi  I have changed them. Thanks for pointing it out.

dkayiwa:
@Michyjones do you know why this is still open?

PR #224, opened 2018-11-14
OCLOMRS-307: Remove unnecessary libraries

AGMETEOR:
# JIRA TICKET NAME:
[Remove unnecessary libraries](https://issues.openmrs.org/browse/OCLOMRS-307)

# Summary:
Remove the following from the application:
- Bootstrap
- Popper JS
- JQuery

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -20,8 +20,5 @@
       You need to enable JavaScript to run this app.
     </noscript>
     <div id="root"></div>
-    <script src="https://code.jquery.com/jquery-3.2.1.slim.min.js" integrity="sha384-KJ3o2DKtIkvYIK3UENzmM7KCkRr/rE9/Qpg6aAZGJwFDMVNA/GpGFF93hXpG5KkN" crossorigin="anonymous"></script>
-    <script src="https://cdnjs.cloudflare.com/ajax/libs/popper.js/1.12.9/umd/popper.min.js" integrity="sha384-ApNbgh9B+Y1QKtv3Rn7W3mgPxhU9K/ScQsAP7hUibX39j7fakFPskvXusvfa0b4Q" crossorigin="anonymous"></script>
-    <script src="https://maxcdn.bootstrapcdn.com/bootstrap/4.0.0/js/bootstrap.min.js" integrity="sha384-JZR6Spejh4U02d8jOt6vLEHfe/JQGiRRSQQxSfFWpi1MquVdAyjUar5+76PVCmYl" crossorigin="anonymous"></script>
   </body>
 </html>


src/components/Navbar.js
```diff
@@ -3,91 +3,86 @@ import Notification, { notify } from 'react-notify-toast';
...
+
+
   logoutUser = (event) => {
     event.preventDefault();
     this.props.logoutAction();
     this.props.history.push('/');
     notify.show('You Loggedout successfully', 'success', 3000);
   };
 
+  toggle() {
```
Duncanian: Kind use an `Arrow function`.

src/components/Navbar.js
```diff
@@ -3,91 +3,86 @@ import Notification, { notify } from 'react-notify-toast';
...
+} from 'reactstrap';
 import { logoutAction } from '../redux/actions/auth/authActions';
 
 export class Navbar extends Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      isOpen: false,
+    };
+    this.toggle = this.toggle.bind(this);
```
Duncanian: This is solved when you use arrow functions

src/components/Navbar.js
AGMETEOR: @Duncanian I have worked on this.

src/components/Navbar.js
AGMETEOR: @Duncanian I have worked on this.

src/tests/Navbar.test.js
```diff
@@ -31,6 +31,19 @@ describe('Navbar Component', () => {
...
+    const props = {
+      loggedIn: true,
+      history: { push: jest.fn() },
+      logoutAction: jest.fn(),
+    };
+    wrapper = shallow(<Navbar.WrappedComponent store={store} {...props} />);
+    const instance = wrapper.instance();
+    expect(wrapper.state().isOpen).toBe(false);
+    instance.toggle();
+    expect(wrapper.state().isOpen).toBe(true);
```
muhozi: @AGMETEOR can you look into this testing block to test properly? like simulating a click on the drop-down.

src/components/Navbar.js
```diff
@@ -3,91 +3,81 @@ import Notification, { notify } from 'react-notify-toast';
...
-                    id="navbarDropdown"
-                    role="button"
-                    data-toggle="dropdown"
-                    aria-haspopup="true"
-                    aria-expanded="false"
-                  >
+        <Navigation color="light" light expand="md">
+          <NavbarBrand href="/">OCL for OpenMRS</NavbarBrand>
+          <NavbarToggler onClick={this.toggle} />
+          {loggedIn ? (
```
dessHub: Use `&&` here instead of `?`.

AGMETEOR: @dessHub I think the ternary makes more sense here. Any particular reason you want && there?

src/tests/Navbar.test.js
muhozi: @AGMETEOR, Adding another test to simulate the click on the dropdown would increase the confidence that our dropdown is working as expected.

AGMETEOR: Let me do that.

src/components/Navbar.js
muhozi: @AGMETEOR I think that in this case using short circuit operator(`&&`) is handier than ternary operator(`?`).

Check this [here](https://reactjs.org/docs/conditional-rendering.html)

src/tests/Navbar.test.js
AGMETEOR: @muhozi I have written a test that takes care of the click.

src/components/Navbar.js
AGMETEOR: Alright. Let me use the short circuit.

AGMETEOR: @dessHub I have made the changes.

dessHub: As @muhozi said `&&` make sense since, you just checking for one condition. `?` is suitable for `if ...else` statements.

dkayiwa:
This pull request has more changes than just "Remove unnecessary libraries"

AGMETEOR:
@dkayiwa The removal of the libraries required a change to some other files that depended on the removed libraries.

dkayiwa:
I not only see removal of libraries, but also change of code.
The two should not be mixed.

pythonmomo:
@dkayiwa alright. Let me separate them and create another ticket.

AGMETEOR:
@dkayiwa let me create a separate ticket

src/tests/Navbar.test.js
```diff
@@ -31,6 +32,30 @@ describe('Navbar Component', () => {
...
+  });
+  it('button click should toggle dropdown', () => {
+    const props = {
```
dondrzzy: Hey @AGMETEOR, seems you are repeating yourself here. The `props` and `shallow rendering`.  It'd be better to define those in a `beforeEach fn` inside the describe block so that its scope is limited to that describe block.

dondrzzy:
@AGMETEOR , this PR also does not seem to properly describe what you are doing. When you say, `remove unnecessary libraries`, the assumption is they are not being used and their removal would not necessitate code changes which you just did so maybe, add another ticket to address the transition from bootstrap to reactstrap and make the changes there or change this ticket's description to that.
*NB* Be certain that the whole application and not just the Navbar component is not depending on the bootstrap css and jquery you intend to omit.

src/tests/Navbar.test.js
AGMETEOR: @dondrzzy it's a good suggestion. However, I would be required to move the previously implemented beforeEach block within the describe block. This I feel is better suited for another ticket that refactors the tests in this file.

AGMETEOR:
@dondrzzy The application has bootstrap that was installed using npm. This makes the bootstrap CDNs redundant.

PR #225, opened 2018-11-16
OCLOMRS-312: Resize the home page margin

Michyjones:
# JIRA TICKET NAME:
[OCLOMRS-312: Resize the home page margin](https://issues.openmrs.org/browse/OCLOMRS-312)

# Summary:
The home page margin is not aligned correctly. The margin should be consistent on the home page and all dictionaries

### DIFF
--- a/src/styles/dashboard.scss
+++ b/src/styles/dashboard.scss
@@ -515,6 +515,9 @@ p{
   margin-left: -42px;
   margin-right: -42px;
 }
+.container{
+  max-width: 1300px;
+}
 
 @media (min-width: 576px){
 .offset-sm-1 {


src/styles/dashboard.scss
```diff
@@ -515,6 +515,9 @@ p{
   margin-left: -42px;
   margin-right: -42px;
 }
+.container{
+  max-width: 1300px;
```
abulojoshua1: @Michyjones, I think it is best if you use percentages instead of pixels here.

dkayiwa:
@Michyjones do you know why this is still open?

PR #231, opened 2018-11-21
OCLOMRS-322: Fix create concept error

ozezzy:
# JIRA TICKET NAME:
[Fix create concept error](https://issues.openmrs.org/browse/OCLOMRS-322)

# Summary:
Fix invalid name error that pops up on creation of a new custom concept, even though fields are correctly filled.


collinewait:
LGTM

src/components/dictionaryConcepts/components/ConceptNameRows.jsx
```diff
@@ -20,7 +20,6 @@ class ConceptNameRows extends Component {
       id: '',
       name: '',
       locale: 'en',
-      locale_full: { value: 'en', label: 'English [en]' },
```
dkayiwa: How did the above line cause the error?

ozezzy: The line did not cause the error but was redundant as the locale_full is fetched using locale.
the error was caused by.

The error was caused by line 22  of
src/components/dictionaryConcepts/components/CreateConceptTable.jsx
where `newRow={{ uuid: newRow }}` over wrote the newRole in default props

dkayiwa: The changes in this pull request should be only and only those required to fix the error as reported on the ticket. If you wanna do some clean up, that should be a separate ticket and hence pull request.

ozezzy: okay, will rectify that.

PR #247, opened 2018-12-03
OCLOMRS-334: Make the link in NotFound.js file to use "Link" instead of "a" tag

collinewait:
# JIRA TICKET NAME:
[Make the link in NotFound.js file to use "Link" instead of "a" tag](https://issues.openmrs.org/browse/OCLOMRS-334)

# Summary:

The Link back to the home page in the NotFound.js file should take advantage of the react-router `Link` rather than HTML's `<a>`

### DIFF
--- a/src/components/NotFound.js
+++ b/src/components/NotFound.js
@@ -1,4 +1,5 @@
 import React from 'react';
+import { Link } from 'react-router-dom';
 
 const NotFound = () => (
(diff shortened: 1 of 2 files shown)


collinewait:
> @collinewait , let the evidence provided be a video instead of a screenshot

@Duncanian, I have added the video.

PR #269, opened 2018-12-11
OCLOMRS-351: Make the Source column and on sidebar show Custom instead of UUID

Duncanian:
# JIRA TICKET NAME:
[Make the Source column and on sidebar show CIEL instead of UUID](https://issues.openmrs.org/browse/OCLOMRS-351)

# Summary:
Source column is wrong. A source should be something like “CIEL”, but what is being displayed in that column looks like a UID. (Also it’s wrong in the sidebar.)

### DIFF
--- a/src/components/dictionaryConcepts/containers/DictionaryConcepts.jsx
+++ b/src/components/dictionaryConcepts/containers/DictionaryConcepts.jsx
@@ -155,6 +155,23 @@ export class DictionaryConcepts extends Component {
     return matchSorter(rows, filter.value, { keys: [id] });
   };
 
+  handleConcepts = (concepts) => {
+    const newConcepts = [];
+    concepts.forEach((concept) => {
+      if (concept.source !== 'CIEL') {
+        const newConcept = {
+          ...concept,
+          source: 'Custom',
+        };
+        newConcepts.push(newConcept);
+      }
+      if (concept.source === 'CIEL') {
+        newConcepts.push(concept);
+      }
+    });
+    return newConcepts;
+  }
+
   render() {
     const {
       match: {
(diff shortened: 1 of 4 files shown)


src/components/dictionaryConcepts/containers/DictionaryConcepts.jsx
```diff
@@ -155,6 +155,23 @@ export class DictionaryConcepts extends Component {
     return matchSorter(rows, filter.value, { keys: [id] });
   };
 
+  handleConcepts = (concepts) => {
```
ozezzy: This code snippet assumes there is only going to be two concept sources ('CIEL' and 'custom'). Can this be modified to generically be concept.source, and only modify known concepts without friendly source name. 
```
concepts.forEach((concept) => {

      if (concept.source === 'something that identifies custom source') {
        const newConcept = {
          ...concept,
          source: 'Custom',
        };
            newConcepts.push(newConcept);
      }else{
           newConcepts.push(concept);
      }
```

Duncanian: @ozezzy , currently in the whole application, a concept is either `CIEL` or `custom`. We have no other sources.

ozezzy: @Duncanian that means this will need to be modified when any other other source is added ?

Duncanian: Yeah, and that will have to be in the post MVP when we will have `PIH concepts`

ozezzy: ok

ozezzy: ok

PR #271, opened 2018-12-13
OCLOMRS-365: Make the modal disappear after deleting a concept and show concepts without the deleted concept.

collinewait:
# JIRA TICKET NAME:
[Make the modal disappear after deleting a concept and show concepts without the deleted concept.](https://issues.openmrs.org/browse/OCLOMRS-365)

# Summary:
Previously, after deleting a concept, the modal did not disappear automatically and after closing it, the table still showed the deleted concept untill the the page was refreshed. That behaviour is fixed in this PR.

### DIFF
--- a/src/redux/reducers/ConceptReducers.js
+++ b/src/redux/reducers/ConceptReducers.js
@@ -89,6 +89,8 @@ export default (state = initialState, action) => {
         ...state,
         paginatedConcepts: state.paginatedConcepts
           .filter(concept => concept.version_url !== action.payload),
+        dictionaryConcepts: state.dictionaryConcepts
+          .filter(concept => concept.version_url !== action.payload),
       };
     case TOTAL_CONCEPT_COUNT:
       return {
--- a/src/tests/dictionaryConcepts/reducer/dictionaryConcept.test.js
+++ b/src/tests/dictionaryConcepts/reducer/dictionaryConcept.test.js
@@ -36,7 +36,7 @@ beforeEach(() => {
   state = {
     concepts: [],
     loading: false,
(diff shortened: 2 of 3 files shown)


src/redux/reducers/ConceptReducers.js
```diff
@@ -89,6 +89,8 @@ export default (state = initialState, action) => {
         ...state,
         paginatedConcepts: state.paginatedConcepts
           .filter(concept => concept.version_url !== action.payload),
+        dictionaryConcepts: state.dictionaryConcepts
```
dkayiwa: Is this part of the ticket?

src/tests/dictionaryConcepts/reducer/dictionaryConcept.test.js
```diff
@@ -36,7 +36,7 @@ beforeEach(() => {
   state = {
     concepts: [],
     loading: false,
-    dictionaryConcepts: [],
+    dictionaryConcepts: [paginatedConcepts.concepts],
```
dkayiwa: Is this part of the ticket?

src/redux/reducers/ConceptReducers.js
collinewait: Yes, it is. In the store `dictionaryConcepts` wasn't updating after deleting a concept and this was the cause of the issue in the ticket.

src/tests/dictionaryConcepts/reducer/dictionaryConcept.test.js
collinewait: Yes, it is. It caters for the updated reducer.

PR #275, opened 2018-12-17
OCLOMRS-361: Disable create button when creating a concept

dessHub:
… new concept

# JIRA TICKET NAME:
[OCLOMRS-361: Disable create button through the process of creating a new concept](https://issues.openmrs.org/browse/OCLOMRS-361)

# Summary:
- Reduced the setTime out as the store was being updated in a shorter time than the set timeout.

### DIFF
--- a/src/components/dictionaryConcepts/containers/CreateConcept.jsx
+++ b/src/components/dictionaryConcepts/containers/CreateConcept.jsx
@@ -93,9 +93,9 @@ export class CreateConcept extends Component {
     const isAddedConcept = addedConcept.length;
     if (isNewConcept && isAddedConcept) {
       setTimeout(() => {
-        notify.show('concept successfully created', 'success', 3000);
         nextProps.history.push(`/concepts/${type}/${typeName}/${collectionName}/${dictionaryName}/${language}`);
-      }, 3000);
+        notify.show('concept successfully created', 'success', 3000);
+      }, 10);
     }
   }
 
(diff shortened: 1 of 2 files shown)


src/components/dictionaryConcepts/containers/CreateConcept.jsx
```diff
@@ -137,7 +137,7 @@ export class CreateConcept extends Component {
     }));
   }
 
-  handleChange(event) {
+   handleChange(event) {
```
collinewait: I think you can revert this change.

dessHub: done

src/components/dictionaryConcepts/containers/CreateConcept.jsx
```diff
@@ -93,9 +93,9 @@ export class CreateConcept extends Component {
     const isAddedConcept = addedConcept.length;
     if (isNewConcept && isAddedConcept) {
       setTimeout(() => {
-        notify.show('concept successfully created', 'success', 3000);
```
dkayiwa: How does this change disable the create button?

dessHub: It doesn't. The reason I did change that was to redirect to the concepts page first, then alert the success message.

dkayiwa: In other words, is this change related to the ticket?

dessHub: Yes, it is. I realized that props were being updated faster than the given timeout hence the button is being re-enabled before the page redirects.

PR #277, opened 2018-12-18
OCLOMRS-369: Hide retired mappings from the list of all dictionary mappings

Duncanian:
# JIRA TICKET NAME:
[Hide retired mappings from the list of all dictionary mappings](https://issues.openmrs.org/browse/OCLOMRS-369)

# Summary:
Currently, after a mapping belonging to a particular dictionary is retired, it's still visible to the user among the list of mappings.

### DIFF
--- a/src/components/dictionaryConcepts/components/ViewConceptMappings.jsx
+++ b/src/components/dictionaryConcepts/components/ViewConceptMappings.jsx
@@ -13,11 +13,22 @@ class ViewConceptMappings extends Component {
     });
   };
 
+  handleMappings = (mappings) => {
+    const newMappings = [];
+    mappings.forEach((mapping) => {
+      if (mapping.retired === false) {
+        newMappings.push(mapping);
+      }
+    });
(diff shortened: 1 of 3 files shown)


src/components/dictionaryConcepts/components/ViewConceptMappings.jsx
```diff
@@ -13,11 +13,22 @@ class ViewConceptMappings extends Component {
...
+    });
+    return newMappings;
+  }
+
   render() {
     const { editModalIsOpen } = this.state;
     const {
       mappings, displayName, showDeleteMappingModal, handleDeleteMapping, source,
     } = this.props;
+    const myMappings = this.handleMappings(mappings);
```
Tittoh: Instead of `myMappings` , you could use a more descriptive name like `activeMappings` or anything appropriate

Duncanian: working on it

PR #287, opened 2019-01-08
OCLOMRS-379: Remove the `No dictionaries found` message when there exists a network error

Mrshinezee:
# JIRA TICKET NAME:
[Remove the `No dictionaries found` message when there exists a network error](https://issues.openmrs.org/browse/OCLOMRS-379)

# Summary:
Whenever there is no internet, and dictionaries are failed to be fetched, a message is returned saying there are no dictionaries, this should be fixed as it can mislead a user that has dictionaries into thinking there are no dictionaries.


src/redux/actions/types.js
```diff
@@ -63,3 +63,4 @@ export const EDIT_CONCEPT_ADD_DESCRIPTION = '[concept] edit_concept_add_descript
 export const EDIT_CONCEPT_REMOVE_ONE_DESCRIPTION = '[concept] edit_concept_remove_one_description';
 export const EDIT_CONCEPT_CREATE_NEW_NAMES = '[concept] edit_concept_create_new_names';
 export const EDIT_CONCEPT_REMOVE_ONE_NAME = '[concept] edit_concept_remove_one_name';
+export const NETWORK_ERROR = '[error] error message';
```
abulojoshua1: @Mrshinezee, I think it is best if you follow the convention being used here of using _ between the words (snake case)

Mrshinezee: done

PR #289, opened 2019-01-08
OCLOMRS-381: show network error message when one tries to search for dictionaries with no internet

Tittoh:
# JIRA TICKET NAME:
[show network error message when one tries to search for dictionaries with no internet](https://issues.openmrs.org/browse/OCLOMRS-381)

# Summary:
Whenever one tries to search for dictionaries when there is a network issue, a loader is shown and no error message is shown in the all dictionaries interface.

### DIFF
--- a/src/redux/actions/dictionaries/dictionaryActionCreators.js
+++ b/src/redux/actions/dictionaries/dictionaryActionCreators.js
@@ -82,18 +82,19 @@ export const fetchDictionaries = () => (dispatch) => {
     });
 };
 
-export const searchDictionaries = searchItem => (dispatch) => {
+export const searchDictionaries = searchItem => async (dispatch) => {
   dispatch(isFetching(true));
-  return api.dictionaries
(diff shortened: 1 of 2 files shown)


src/redux/actions/dictionaries/dictionaryActionCreators.js
```diff
@@ -83,17 +83,22 @@ export const fetchDictionaries = () => (dispatch) => {
...
-    .searchDictionaries(searchItem)
-    .then((payload) => {
-      dispatch(isSuccess(payload));
-      dispatch(isFetching(false));
-    })
-    .catch(error => () => {
-      dispatch(isErrored(error.payload.data));
-      dispatch(isFetching(false));
-    });
+  if (navigator.onLine){
```
collinewait: `navigator.onLine` will be `true` so long as the machine is connected to a `network` even in the absence of `internet`. Won't it give us false positives?

PR #293, opened 2019-01-14
OCLOMRS-384: Fix warnings on browser and terminal console logs

collinewait:
# JIRA TICKET NAME:
[Fix warnings on the browser and terminal console logs](https://issues.openmrs.org/browse/OCLOMRS-384)

# Summary:
There are errors on the terminal console log when running tests and also a few that appear in the browsers console logs. These can cause issues during production if not caught early.

### DIFF
--- a/src/components/bulkConcepts/component/CardBody.jsx
+++ b/src/components/bulkConcepts/component/CardBody.jsx
@@ -10,7 +10,11 @@ const CardBody = ({ title, body }) => (
 
 CardBody.propTypes = {
   title: PropTypes.string.isRequired,
-  body: PropTypes.string.isRequired,
+  body: PropTypes.string,
+};
+
+CardBody.defaultProps = {
+  body: null,
 };
 
 export default CardBody;
--- a/src/components/bulkConcepts/component/PreviewCard.jsx
+++ b/src/components/bulkConcepts/component/PreviewCard.jsx
@@ -31,7 +31,7 @@ const PreviewCard = ({
     minute: 'numeric',
   };
 
-  const mapping = mappings ? mappings.length : 'none';
+  const mapping = mappings || 'none';
   const description = descriptions ? descriptions[0].description : 'none';
   return (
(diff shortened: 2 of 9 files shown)


dkayiwa:
@collinewait why hasn't this got a review for 4 days?

collinewait:
> @collinewait why hasn't this got a review for 4 days?

It was ready by Monday but new issues were introduced after merging some PRs and I am fixing them.

src/components/bulkConcepts/component/CardBody.jsx
```diff
@@ -10,7 +10,11 @@ const CardBody = ({ title, body }) => (
 
 CardBody.propTypes = {
   title: PropTypes.string.isRequired,
-  body: PropTypes.string.isRequired,
+  body: PropTypes.string,
```
dessHub: Is `body` not required?

collinewait: When `Add CIEL Concepts` is being clicked, `CardBody` will be called before concepts are in the store. Making it required causes a warning since props passed to `CardBody` are obtained from the store.

src/components/bulkConcepts/component/PreviewCard.jsx
```diff
@@ -31,7 +31,7 @@ const PreviewCard = ({
     minute: 'numeric',
   };
 
-  const mapping = mappings ? mappings.length : 'none';
```
GodswillOnuoha: Is changing this line necessary ?

collinewait: Yeah. `MappingPreview` component  being called here`<MappingPreview title="Mappings" body={mapping} />` is supposed to show `mappings` but `const mapping = mappings ? mappings.length : 'none';` only sends the `number` of mappings available on a concept and `NOT` the mappings themselves. It was also causing an error because the react table in `MappingPreview` was not expecting a number.

PR #296, opened 2019-01-17
OCLOMRS-383: add more sources and source search

ozezzy:
# JIRA TICKET NAME:
[add more selectable concept sources to addBulk concept page](https://issues.openmrs.org/browse/OCLOMRS-383)

# Summary:
user should be able to select a source other than CIEL when adding bulk concepts. there should be a search field to search for and select a source not listed as part of the radio buttons.


src/components/bulkConcepts/addBulkConcepts.jsx
```diff
@@ -35,24 +37,44 @@ export class AddBulkConcepts extends Component {
...
     };
     this.invalidConceptIds = [];
-    this.sourceUrl = 'orgs/CIEL/sources/CIEL/concepts/';
+    this.sourceUrl = 'orgs/CIEL/sources/CIEL/';
   }
 
   componentDidMount() {
-    this.props.fetchCielConcepts();
+    this.props.getConceptSources();
+    this.props.fetchSourceConcepts(this.sourceUrl);
```
abulojoshua1: @ozezzy, kindly destructure props here.

src/components/bulkConcepts/addBulkConcepts.jsx
```diff
@@ -35,24 +37,44 @@ export class AddBulkConcepts extends Component {
...
+  handleCielClick = (e) => {
+    this.setState({ otherSelected: false });
+    this.sourceUrl = e.target.value;
+    this.props.fetchSourceConcepts(this.sourceUrl);
+    this.reset();
   };
 
+  otherSourceClick = () => {
+    this.setState({ otherSelected: true });
+    this.props.getConceptSources();
```
abulojoshua1: @ozezzy, kindly destructure props here.

src/components/bulkConcepts/addBulkConcepts.jsx
```diff
@@ -35,24 +37,44 @@ export class AddBulkConcepts extends Component {
...
+    this.reset();
+  }
+
+  handleSourceSelect = (source) => {
+    this.sourceUrl = source.url.substring(1);
+    this.props.fetchSourceConcepts(this.sourceUrl);
```
abulojoshua1: @ozezzy, kindly destructure props here.

Tittoh:
@ozezzy Since we already have the CIEL radio button, do we need to have it in the dropdown too?

src/components/bulkConcepts/addBulkConcepts.jsx
ozezzy: Destructuring fetchSourceConcepts here will give a linting error

src/components/bulkConcepts/addBulkConcepts.jsx
ozezzy: Destructuring getConceptSources and fetchSourcesConcepts here gives linting error because of line 11. They are already declared in the scope.

src/components/bulkConcepts/addBulkConcepts.jsx
ozezzy: destructuring  getConceptSources here give linting error

ozezzy:
> @ozezzy Since we already have the CIEL radio button, do we need to have it in the dropdown too?

CIEL is just added as a radio button for quick access, I don't see why to remove it from the list of available sources returned from the API (the dropdown). Let me know if removing it gives better usability though

src/components/bulkConcepts/addBulkConcepts.jsx
```diff
@@ -35,24 +37,44 @@ export class AddBulkConcepts extends Component {
     this.state = {
       conceptIds: '',
       openResultModal: false,
+      otherSelected: false,
     };
     this.invalidConceptIds = [];
-    this.sourceUrl = 'orgs/CIEL/sources/CIEL/concepts/';
+    this.sourceUrl = 'orgs/CIEL/sources/CIEL/';
```
dkayiwa: What has this got to do with other sources other than CIEL?

src/components/bulkConcepts/addBulkConcepts.jsx
```diff
@@ -129,7 +152,7 @@ Dictionary
                 type="radio"
                 name="exampleRadios"
                 id="ciel"
-                value="option1"
+                value="/orgs/CIEL/sources/CIEL/"
```
dkayiwa: What has this got to do with other sources other than CIEL?

dkayiwa:
The changes in this pull request seem to be doing more than a simple "add more selectable concept sources to addBulk concept page"

src/components/bulkConcepts/addBulkConcepts.jsx
ozezzy: CIEL is the default when the page is loaded. `this.sourceUrl = 'orgs/CIEL/sources/CIEL/'` holds the sourceUrl, and it changes when a different source is selected.

src/components/bulkConcepts/addBulkConcepts.jsx
ozezzy: from my comment above, when a different source is selected, CIEL radio button is deselected, and `this.sourceUrl` changes to the url of the selected source. If CIEL radio button is selected agin, it gets the  the value `value="/orgs/CIEL/sources/CIEL/"` and assigns it to `this.sourceUrl` again.

ozezzy:
It only added auto suggestion while searching on other sources, and modified the implementation of CIEL only format, to enable it work properly and reuse functions such as `validConcepts` lines 96 and 98 of `src/components/bulkConcepts/addBulkConcepts.jsx`

dkayiwa:
All changes in this pull request should simply be about ""add more selectable concept sources to addBulk concept page". Any other fixes or improvements should be another ticket and hence pull request.

ozezzy:
Ok. I will look through this, and see if there are changes not necessary in this PR

ozezzy:
I Think I better understand why it appears like there are unneeded changes. I made changes to initial state,  action and reducer functions that handled CIEL concept source to accommodate for other concept sources. I think this is justifiable in this case, as opposed to duplicating the set of action and reducer functions of the CIEL source implementation and changing the source URL.  Every change here is needed for this implementation to work. However, If this is not acceptable, I could close this PR and use the second approach. The resulting code would be much different from the current.

dkayiwa:
What would be the second approach?

ozezzy:
to create different sets of action and reducer functions, generally try to make minimal changes to the way it currently works for CIEL source. And then create a different ticket for improving the code

dkayiwa:
Which do you think is the better approach?

ozezzy:
I think the current approach is good in this case, It saves time, and the changes are all needed for the feature to work as implemented. Also considering the resulting code isn't too complex to reason along with.

dkayiwa:
In your video, you choose other and then select CIEL. Does this make sense when you already have CIEL up?

src/components/bulkConcepts/addBulkConcepts.jsx
```diff
@@ -138,6 +161,84 @@ Dictionary
...
             </div>
             <br />
+
+
+            <div id="other-search">
+              <div className="form-check">
+                <input
+                  className="form-check-input"
+                  type="radio"
+                  name="exampleRadios"
```
dkayiwa: Did you intentionally call it exampleRadios?

src/components/bulkConcepts/addBulkConcepts.jsx
```diff
@@ -138,6 +161,84 @@ Dictionary
...
+                  id="otherSourcesOption"
+                  onClick={this.otherSourceClick}
+                  value="option3"
```
dkayiwa: Can we get a better value other than option3?

src/components/bulkConcepts/addBulkConcepts.jsx
```diff
@@ -138,6 +161,84 @@ Dictionary
...
+                            || item.name.toLowerCase()
+                              .includes(inputValue.toLowerCase())).slice(1, 10)
+                              .map((item, index) => (
+                                <li
+                                  {...getItemProps({
+                                    key: item.id,
+                                    index,
+                                    item,
+                                    style: {
+                                      backgroundColor:
```
dkayiwa: Can't we do this from styling file?

src/components/bulkConcepts/addBulkConcepts.jsx
ozezzy: no, it is now fixed.

src/components/bulkConcepts/addBulkConcepts.jsx
ozezzy: I have removed 'value'. it is not needed

src/components/bulkConcepts/addBulkConcepts.jsx
ozezzy: No, it is passed as props to downshift directly, and contains logic

ozezzy:
> In your video, you choose other and then select CIEL. Does this make sense when you already have CIEL up?

I have fixed this and replaced the video

src/components/bulkConcepts/addBulkConcepts.jsx
```diff
@@ -8,17 +8,19 @@ import {
   FormGroup,
   Input,
 } from 'reactstrap';
-import fetchCielConcepts, { addExistingBulkConcepts, isConceptValid } from '../../redux/actions/bulkConcepts';
+import fetchSourceConcepts, { addExistingBulkConcepts, isConceptValid, getConceptSources } from '../../redux/actions/bulkConcepts';
```
dkayiwa: Shouldn't the above be named fetchConceptSources

src/components/bulkConcepts/addBulkConcepts.jsx
```diff
@@ -138,6 +161,84 @@ Dictionary
...
+                      <div id="other-search">
+                          Other &nbsp;&nbsp;
+                        <form className="form-inline search-bar">
+                          <i className="fas fa-search" />
+                          { otherSelected && <input
+                            {
+                            ...getInputProps()
+                            }
+                            className="form-control search"
+                            id="sourceSearch"
```
dkayiwa: Is the placeholder search with small s?

src/redux/actions/bulkConcepts/index.js
```diff
@@ -1,21 +1,31 @@
...
+} from '../globalActionCreators';
+import {
+  FETCH_SOURCE_CONCEPTS,
+  ADD_EXISTING_BULK_CONCEPTS,
+  FETCH_CONCEPT_SOURCES,
+  CLEAR_SOURCE_CONCEPTS,
+} from '../types';
 
-const fetchCielConcepts = () => async (dispatch) => {
+const fetchSourceConcepts = (url = 'orgs/CIEL/sources/CIEL/') => async (dispatch) => {
```
dkayiwa: Are we hard coding this url value?

src/components/bulkConcepts/addBulkConcepts.jsx
ozezzy: fixed

src/components/bulkConcepts/addBulkConcepts.jsx
ozezzy: fixed

src/redux/actions/bulkConcepts/index.js
ozezzy: fixed

PR #304, opened 2019-01-22
OCLOMRS-402: The buttons theme for (Subscription URL popup, Release a new version popup, Confirm release popup) is inconsistent

judeinno:
# JIRA TICKET NAME:
[Make buttons consistent](https://issues.openmrs.org/browse/OCLOMRS-402)

# Summary:
The buttons theme for (Subscription URL popup, Release a new version popup, Confirm release popup) is inconsistent
- The buttons on these popups do not look like the other buttons in the application hence the inconsistency

### DIFF
--- a/src/styles/dictionary_modal.scss
+++ b/src/styles/dictionary_modal.scss
@@ -147,11 +147,7 @@ legend {
       color: #cc0000;
   }
 
-button:disabled{
-    cursor: not-allowed;
-    background: #e9ecef !important;
-}
-#addDictionary {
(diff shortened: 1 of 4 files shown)


src/styles/dictionary_modal.scss
```diff
@@ -147,11 +147,7 @@ legend {
       color: #cc0000;
   }
 
-button:disabled{
```
AGMETEOR: @judeinno just asking. Won't this style be needed somewhere else in the application?

judeinno: We are using the default disabled for consistency.

PR #307, opened 2019-01-23
OCLOMRS-394: Add a loader to addBulk concepts search

Mrshinezee:
# JIRA TICKET NAME:
[Add a loader to add bulk concepts search](https://issues.openmrs.org/browse/OCLOMRS-394)

# Summary:
A user should see the loader to show something is going on while the search is loading

### DIFF
--- a/src/components/Loader.js
+++ b/src/components/Loader.js
@@ -1,7 +1,16 @@
 import React from 'react';
+import PropTypes from 'prop-types';
 
-const Loader = () => (
-  <div className="loader" />
+const Loader = ({ smaller }) => (
+  <div className={`${smaller ? 'smaller-loader' : 'loader'}`} />
 );
+
+Loader.propTypes = {
+  smaller: PropTypes.bool,
+};
+
+Loader.defaultProps = {
+  smaller: false,
+};
 
 export default Loader;
--- a/src/components/bulkConcepts/addBulkConcepts.jsx
+++ b/src/components/bulkConcepts/addBulkConcepts.jsx
@@ -8,6 +8,7 @@ import {
   FormGroup,
   Input,
 } from 'reactstrap';
+import Loader from '../Loader';
 import fetchSourceConcepts, { addExistingBulkConcepts, isConceptValid, fetchConceptSources } from '../../redux/actions/bulkConcepts';
 import Header from './container/Header';
 import ResultModal from './component/addBulkConceptResultModal';
@@ -30,6 +31,7 @@ export class AddBulkConcepts extends Component {
       }).isRequired,
     }).isRequired,
     isFetching: PropTypes.bool.isRequired,
+    isLoading: PropTypes.bool.isRequired,
   };
 
   constructor(props) {
@@ -123,6 +125,8 @@ export class AddBulkConcepts extends Component {
       },
       sourceConcepts,
       conceptSources,
+      isLoading,
+      isFetching,
     } = this.props;
     const disableButton = (conceptIds.length < 1);
     return (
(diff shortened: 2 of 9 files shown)


src/components/Loader.js
```diff
@@ -1,7 +1,16 @@
 import React from 'react';
+import PropTypes from 'prop-types';
 
-const Loader = () => (
-  <div className="loader" />
+const Loader = ({ smaller }) => (
```
collinewait: Are all loaders in the system made smaller?

Mrshinezee: No, so by default you will have the loader with classname loader. but in case you want a smaller loader like I did. You will need to pass a prop of smaller to the loader.

src/redux/actions/bulkConcepts/index.js
```diff
@@ -11,18 +11,19 @@ import {
   ADD_EXISTING_BULK_CONCEPTS,
   FETCH_CONCEPT_SOURCES,
   CLEAR_SOURCE_CONCEPTS,
+  IS_LOADING,
 } from '../types';
 
 const fetchSourceConcepts = url => async (dispatch) => {
   dispatch(clear(CLEAR_SOURCE_CONCEPTS));
-  dispatch(isFetching(true));
+  dispatch(isSuccess(true, IS_LOADING));
```
GodswillOnuoha: are the changes in this file necessary? . dispatching ` isFetching(true/false) ` should set `loading status` to true of false, and that can be used

Mrshinezee: Yes, it is. There are two calls made one with the fetchSourceConcepts and the other was fetchConceptSource. Both do set loading to false once completed. There was a need to differentiate them. That was why I changed it a different name.

src/components/bulkConcepts/addBulkConcepts.jsx
```diff
@@ -189,43 +193,48 @@ Dictionary
...
+                                <Loader smaller />
+                              </div>
+                            }
                         </form>
                       </div>
                       <div className="search-ul">
                         {isOpen ? (
                           <ul {...getMenuProps()} className="search-ul">
                             {
-                            conceptSources.filter(item => !inputValue.trim()
```
dkayiwa: Is this just a formatting change?

src/components/bulkConcepts/addBulkConcepts.jsx
```diff
@@ -261,44 +270,50 @@ Dictionary
...
+                            }
                         </form>
                       </div>
                     </div>
                     <div className="search-ul">
                       {isOpen ? (
                         <ul {...getMenuProps()} className="search-ul">
                           {
-                          sourceConcepts.filter(item => !inputValue.trim()
-                          || item.display_name.toLowerCase()
```
dkayiwa: Is this just a formatting change?

src/components/bulkConcepts/addBulkConcepts.jsx
Mrshinezee: Yes, It was because I added the div for the loader component

src/components/bulkConcepts/addBulkConcepts.jsx
Mrshinezee: Yes, It was because I added the div for the loader component

Mrshinezee: Yes it was on the search input
```
                     {isLoading
                              && <div className="ml-auto text-right">
                                <Loader smaller />
                              </div>
                            }
```

PR #308, opened 2019-01-24
OCLOMRS-397: Fix inaccurate total concept counts on some dictionary detail card page

AGMETEOR:
# JIRA TICKET NAME:
[Fix inaccurate total concept counts on dictionary detail card page](https://issues.openmrs.org/browse/OCLOMRS-397)

# Summary:
On the dictionary overview page, the dictionary overview card shows total concepts count that doesn't match the sum between custom concepts and CIEL concepts. With some dictionaries this sum is fine and the total concepts count is representative, however, some like [this one](https://openmrs.qa.openconceptlab.org/dictionaryOverview/users/admin/collections/ME101/) show this discrepancy.

The issue is that active_concepts field that is sent with the dictionary from the backend is used to represent total concepts on the frontend. This is not right.

PR #310, opened 2019-01-25
OCLOMRS-409: Add display for active concepts on the dictionary detail card

AGMETEOR:
# JIRA TICKET NAME:
[Add display for active concepts on the dictionary detail card](https://issues.openmrs.org/browse/OCLOMRS-409)

# Summary:
In order to tell the user what type of concepts they are looking at, there's a need to also add a display that explicitly tells the user the active concepts count on the dictionary detail card. This ticket is based on feedback from this [thread.](https://talk.openmrs.org/t/developing-the-ocl-for-openmrs-application/17771/537)

### DIFF
--- a/src/components/dashboard/components/dictionary/DictionaryDetailCard.jsx
+++ b/src/components/dashboard/components/dictionary/DictionaryDetailCard.jsx
@@ -144,6 +144,12 @@ Dictionary
               {active_concepts}
             </p>
             <p className="points">
+              <b>Active Concepts</b>
+:
+              {' '}
+              {active_concepts}
+            </p>
+            <p className="points">
               <b>Custom Concepts</b>
 :
               {' '}


src/components/dashboard/components/dictionary/DictionaryDetailCard.jsx
```diff
@@ -139,6 +139,12 @@ Dictionary
             <legend>Concepts (HEAD version)</legend>
             <p className="paragraph">
               <b>Total Concepts</b>
+:
+              {' '}
+              {active_concepts}
```
brucemakallan: Is `active_concepts` the same as total concepts?
I've noticed that it is repeated below. Shouldn't we just keep one if they are the same?

AGMETEOR: @brucemakallan no it's not the same. There's another [PR](https://github.com/openmrs/openmrs-ocl-client/pull/308) that fixes that issue.

brucemakallan: OK. Thanks

PR #316, opened 2019-01-29
OCLOMRS-412: Make Datatypes and Classes sidebar filters work as expected

collinewait:
# JIRA TICKET NAME:
[Make Datatypes and Classes sidebar filters work as expected](https://issues.openmrs.org/browse/OCLOMRS-412)

# Summary:
On Add CIEL concepts page when one tries to filter the results with Datatypes and Classes sidebar filters, nothing happens and an error is displayed.

### DIFF
--- a/src/components/bulkConcepts/container/BulkConceptsPage.jsx
+++ b/src/components/bulkConcepts/container/BulkConceptsPage.jsx
@@ -78,10 +78,12 @@ export class BulkConceptsPage extends Component {
     } = event;
     const targetName = name.split(',');
     const query = `q=${this.state.searchInput}`;
+    this.setState({ searchingOn: true });
+    const { currentPage } = this.props;
     if ((targetName[1]).trim() === 'datatype') {
-      this.props.addToFilterList(targetName[0], 'datatype', query);
+      this.props.addToFilterList(targetName[0], 'datatype', query, currentPage);
     } else {
-      this.props.addToFilterList(targetName[0], 'classes', query);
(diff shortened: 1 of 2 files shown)


src/components/bulkConcepts/container/BulkConceptsPage.jsx
```diff
@@ -78,10 +78,12 @@ export class BulkConceptsPage extends Component {
     } = event;
     const targetName = name.split(',');
     const query = `q=${this.state.searchInput}`;
+    this.setState({ searchingOn: true });
+    const { currentPage } = this.props;
     if ((targetName[1]).trim() === 'datatype') {
-      this.props.addToFilterList(targetName[0], 'datatype', query);
+      this.props.addToFilterList(targetName[0], 'datatype', query, currentPage);
```
brucemakallan: Since `addToFilterList` is reused, Please add it to the destructuring on line `82`

collinewait: Destructuring the action `addToFilterList` on line `82` causes `no-shadow` ESlint error. There are workarounds that can be made as observed [here](https://stackoverflow.com/questions/37682705/avoid-no-shadow-eslint-error-with-mapdispatchtoprops) but they can introduce in some complexity like reducing readability, introducing in some boilerplate code and disabling eslint isn't ideal. By explicitly using the method off of the props object, we don't need to worry about shadowing.

brucemakallan: Alright

PR #318, opened 2019-01-30
OCLOMRS-410: Enable Partial word searching on dictionary concepts page

brucemakallan:
# JIRA TICKET NAME:
[Enable Partial word searching on dictionary concepts page](https://issues.openmrs.org/browse/OCLOMRS-410)

# Summary:
Enable users to search for concepts by entering one or more letters as opposed to complete words only.

### DIFF
--- a/src/components/bulkConcepts/container/BulkConceptsPage.jsx
+++ b/src/components/bulkConcepts/container/BulkConceptsPage.jsx
@@ -1,6 +1,7 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
+import { NotificationContainer, NotificationManager } from 'react-notifications';
 import SideNav from '../component/Sidenav';
 import SearchBar from '../component/SearchBar';
 import ConceptTable from '../component/ConceptTable';
@@ -14,6 +15,8 @@ import {
   setCurrentPage,
 } from '../../../redux/actions/concepts/addBulkConcepts';
 import { conceptsProps } from '../../dictionaryConcepts/proptypes';
+import { MIN_CHARACTERS_WARNING, MILLISECONDS_TO_SHOW_WARNING } from '../../../redux/reducers/generalSearchReducer';
+import { INTERNAL_MAPPING_DEFAULT_SOURCE } from '../../dictionaryConcepts/components/helperFunction';
 
 export class BulkConceptsPage extends Component {
   static propTypes = {
@@ -105,9 +108,11 @@ export class BulkConceptsPage extends Component {
 
   searchOption = () => {
     const { searchInput } = this.state;
-    const query = `q=${searchInput.trim()}`;
-    const { currentPage, fetchFilteredConcepts: fetchedFilteredConcepts } = this.props;
-    fetchedFilteredConcepts('CIEL', query, currentPage);
+    if (searchInput && searchInput.trim().length > 2) {
+      const query = `q=${searchInput.trim()}*`; // The asterisk permits partial search
+      const { currentPage, fetchFilteredConcepts: fetchedFilteredConcepts } = this.props;
+      fetchedFilteredConcepts(INTERNAL_MAPPING_DEFAULT_SOURCE, query, currentPage);
+    } else NotificationManager.warning(MIN_CHARACTERS_WARNING, '', MILLISECONDS_TO_SHOW_WARNING);
   }
 
   handleSearch = async (event) => {
@@ -166,6 +171,7 @@ export class BulkConceptsPage extends Component {
               handleChange={this.handleChange}
               searchInput={searchInput}
             />
+            <NotificationContainer />
             <ConceptTable
               concepts={concepts}
               loading={loading}
--- a/src/components/dashboard/container/DictionariesDisplay.jsx
+++ b/src/components/dashboard/container/DictionariesDisplay.jsx
@@ -2,6 +2,7 @@ import React, { Component, Fragment } from 'react';
 import autoBind from 'react-autobind';
 import propTypes from 'prop-types';
 import { connect } from 'react-redux';
+import { NotificationContainer, NotificationManager } from 'react-notifications';
 import {
   fetchDictionaries,
   searchDictionaries,
@@ -11,6 +12,7 @@ import ListDictionaries from '../components/dictionary/ListDictionaries';
 import SearchDictionaries from '../components/dictionary/DictionariesSearch';
 import Paginations from '../components/DictionariesPaginations';
 import Title from '../../Title';
+import { MIN_CHARACTERS_WARNING, MILLISECONDS_TO_SHOW_WARNING } from '../../../redux/reducers/generalSearchReducer';
 
 export class DictionaryDisplay extends Component {
   static propTypes = {
@@ -40,10 +42,13 @@ export class DictionaryDisplay extends Component {
     this.props.fetchDictionaries();
   }
 
-  onSubmit(event) {
+  onSubmit = (event) => {
     event.preventDefault();
     this.props.clearDictionaries();
-    this.props.searchDictionaries(this.state.searchInput);
+    const { searchInput } = this.state;
+    if (searchInput && searchInput.trim().length > 2) {
+      this.props.searchDictionaries(`${searchInput}*`); // asterisk added to allow partial search
+    } else NotificationManager.warning(MIN_CHARACTERS_WARNING, '', MILLISECONDS_TO_SHOW_WARNING);
   }
 
   onSearch(event) {
@@ -91,6 +96,7 @@ export class DictionaryDisplay extends Component {
               searchValue={searchInput}
               fetching={isFetching}
             />
+            <NotificationContainer />
           </div>
         </div>
         <Paginations
--- a/src/redux/reducers/generalSearchReducer.js
+++ b/src/redux/reducers/generalSearchReducer.js
@@ -1,5 +1,8 @@
 import { SEARCH_RESULTS, IS_FETCHING } from '../actions/types';
 
+export const MIN_CHARACTERS_WARNING = 'Search query must contain at least 3 characters';
+export const MILLISECONDS_TO_SHOW_WARNING = 4000;
+
 const initialState = { dictionaries: [], loading: false };
 
 export default (state = initialState, action) => {
--- a/src/tests/Dashboard/action/dictionaryAction.test.js
+++ b/src/tests/Dashboard/action/dictionaryAction.test.js
@@ -38,7 +38,7 @@ import {
   createVersion,
   editMapping,
 } from '../../../redux/actions/dictionaries/dictionaryActionCreators';
-import dictionaries from '../../__mocks__/dictionaries';
+import dictionaries, { sampleDictionaries } from '../../__mocks__/dictionaries';
 import versions, { HeadVersion } from '../../__mocks__/versions';
 import concepts from '../../__mocks__/concepts';
 
@@ -266,13 +266,13 @@ describe('Test suite for dictionary actions', () => {
       const request = moxios.requests.mostRecent();
       request.respondWith({
         status: 200,
-        response: [dictionaries],
+        response: sampleDictionaries,
       });
     });
 
     const expectedActions = [
       { type: IS_FETCHING, payload: true },
-      { type: FETCHING_DICTIONARIES, payload: [dictionaries] },
+      { type: FETCHING_DICTIONARIES, payload: sampleDictionaries },
       { type: IS_FETCHING, payload: false },
     ];
 
--- a/src/tests/Dashboard/container/DictionaryDisplay.test.jsx
+++ b/src/tests/Dashboard/container/DictionaryDisplay.test.jsx
@@ -104,28 +104,38 @@ describe('DictionaryDisplay', () => {
     expect(mapStateToProps(initialState).isFetching).toEqual(false);
   });
   it('should search for a dictionary', () => {
+    const searchDictionaries = jest.fn();
+    const clearDictionaries = jest.fn();
     const props = {
       fetchDictionaries: jest.fn(),
       dictionaries: [],
       isFetching: true,
-      clearDictionaries: jest.fn(),
-      searchDictionaries: jest.fn(),
+      clearDictionaries,
+      searchDictionaries,
       onSearch: jest.fn(),
       onSubmit: jest.fn(),
       organizations: [organizations],
       history: { push: jest.fn() },
-      searchValue: '',
     };
-    const wrapper = mount(<MemoryRouter>
-      <DictionaryDisplay {...props} />
-    </MemoryRouter>);
+    const wrapper = mount(<DictionaryDisplay {...props} />);
     const event = { target: { name: 'searchInput', value: 'openmrs' } };
     const event2 = { target: { name: 'searchInput', value: '', type: 'checkbox' } };
     wrapper.find('#search').simulate('change', event);
     wrapper.find('#search').simulate('change', event2);
-    wrapper.find('#submit-search-form').simulate('submit', {
-      preventDefault: () => {},
+    const onSubmitEvent = { preventDefault: jest.fn() };
+    wrapper.setState({ searchInput: 'ka' }, () => {
+      wrapper.update();
+      wrapper.find('#submit-search-form').simulate('submit', onSubmitEvent);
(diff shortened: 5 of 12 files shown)


src/components/bulkConcepts/component/SearchBar.jsx
```diff
@@ -17,6 +17,8 @@ const SearchBar = ({
           onChange={handleChange}
           placeholder="search concept"
           required
+          pattern="^.{3,}$"
```
dkayiwa: Can we avoid duplicating the pattern and message?

brucemakallan: @dkayiwa I have declared the pattern and the message/warning in the `generalSearchReducer` and passed it through props.
I've also provided the required mock data for all tests that required those new attributes.

brucemakallan: @dkayiwa I would have to change it in just one place actually. All those other times are just mock data for tests. I found that the way tests were written, mock data must be passed for all tests that require it. That means all tests that have Dictionary Search or Concepts Search. 

*For example:* Mock data is repeated (as props) for tests within this file: `https://github.com/openmrs/openmrs-ocl-client/blob/master/src/tests/Dashboard/container/DictionaryDisplay.test.jsx`

dkayiwa: If you changed it in that one place, would the tests still be testing 4 characters or 3?

brucemakallan: @dkayiwa I understand the concern and have therefore created a single source of truth for tests as well. The pattern and the message are now declared only ONCE in the entire application.

src/redux/reducers/generalSearchReducer.js
```diff
@@ -1,6 +1,14 @@
 import { SEARCH_RESULTS, IS_FETCHING } from '../actions/types';
 
-const initialState = { dictionaries: [], loading: false };
+export const pattern = '^.{3,}$';
```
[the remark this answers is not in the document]
brucemakallan: @dkayiwa I looked at the ES6 naming conventions for constants (https://github.com/airbnb/javascript#naming--uppercase) and have made the appropriate changes.

dkayiwa:
Instead of saying that please match the requested format, why don't you just give the three minimum characters message when one attempts to search?

dkayiwa:
Do you then need the regex pattern?

brucemakallan:
> Do you then need the regex pattern?

@dkayiwa Yes. Because it's the one that triggers the `onInvalid` method

dkayiwa:
I thought you would check the character length only at the time when the user initiates the search.

brucemakallan:
> I thought you would check the character length only at the time when the user initiates the search.

@dkayiwa True. And it runs the check for the pattern. If all is well, it proceeds with the search. If not, it displays the message.

dkayiwa:
Do you need the complexity of the pattern? Can't you just check the character length?

brucemakallan:
> Do you need the complexity of the pattern? Can't you just check the character length?

Problem is that without a regex pattern provided to the JSX input field, the `onInvalid` method won't be triggered. Meaning that I won't be able to override its behaviour (to provide a custom message format).

dkayiwa:
I mean just giving the error message when once clicks the search button.

brucemakallan:
> I mean just giving the error message when once clicks the search button.

That error message will have a heading of *"Please match the requested format"*. That's why I needed an override of the default behaviour.

dkayiwa:
What i mean is still. When one clicks the search button, can't you get the input length and then give a message of the three character minimum requirement?

brucemakallan:
> What i mean is still. When one clicks the search button, can't you get the input length and then give a message of the three character minimum requirement?

I can. But I would have to use my own alert messaging system or a third party library for alerts. I had wanted to stick to the one provided by HTML5.

brucemakallan:
Would you rather I went with that implementation instead?

src/tests/bulkConcepts/containers/BulkConceptsPage.test.js
```diff
@@ -537,11 +565,68 @@ describe('Test suite for BulkConceptsPage component', () => {
...
+    const setCustomValidity = jest.fn();
+    const event = {
+      target: {
+        setCustomValidity,
+        value: 'ka',
+      },
+    };
+    wrapper.find('input#search-concept').simulate('invalid', event);
+    expect(setCustomValidity).toHaveBeenCalled();
+    // if search query >= 3 characters
```
dkayiwa: What is the above comment for?

src/tests/bulkConcepts/containers/BulkConceptsPage.test.js
```diff
@@ -537,11 +565,68 @@ describe('Test suite for BulkConceptsPage component', () => {
...
+      datatypes: [],
+      classes: [],
+      conceptLimit: 10,
+      match: {
+        params: {
+          type: 'users',
+          typeName: 'emasys',
+          collectionName: 'dev-org',
+          language: 'en',
+          dictionaryName: 'CIEL',
```
dkayiwa: Don't we already have a constant for CIEL?

src/tests/Dashboard/container/DictionaryDisplay.test.jsx
```diff
@@ -133,10 +150,47 @@ describe('DictionaryDisplay', () => {
...
+    const setCustomValidity = jest.fn();
+    const event = {
+      target: {
+        setCustomValidity,
+        value: 'mu',
+      },
+    };
+    wrapper.find('input#search').simulate('invalid', event);
+    expect(setCustomValidity).toHaveBeenCalled();
+    // if search query >= 3 characters
```
dkayiwa: What is the above comment for?

brucemakallan:
> @dkayiwa Thank you for the reviews and the comments. I am rewriting the entire implementation to go with what you suggested in the hopes that it will reduce the unnecessary complexity. I'll reach out as soon as it is ready.

@dkayiwa As promised earlier, I have redone the task and updated the video on JIRA as well.

src/components/bulkConcepts/container/BulkConceptsPage.jsx
```diff
@@ -102,9 +104,11 @@ export class BulkConceptsPage extends Component {
 
   searchOption = () => {
     const { searchInput } = this.state;
-    const query = `q=${searchInput.trim()}`;
-    const { currentPage, fetchFilteredConcepts: fetchedFilteredConcepts } = this.props;
-    fetchedFilteredConcepts('CIEL', query, currentPage);
+    if (searchInput && searchInput.trim().length > 2) {
+      const query = `q=${searchInput.trim()}*`; // The asterisk permits partial search
+      const { currentPage, fetchFilteredConcepts: fetchedFilteredConcepts } = this.props;
+      fetchedFilteredConcepts('CIEL', query, currentPage);
```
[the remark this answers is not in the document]
brucemakallan: @dkayiwa I've just searched through the files and didn't find a constant for "CIEL". I left the code on line 110 as the author had written it. I only added the `if ... else` logic.

src/redux/reducers/generalSearchReducer.js
```diff
@@ -1,5 +1,7 @@
 import { SEARCH_RESULTS, IS_FETCHING } from '../actions/types';
 
+export const WARNING = 'Search query must contain at least 3 characters';
```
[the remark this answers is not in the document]
brucemakallan: @dkayiwa The constant "WARNING" represents the message that the user will see if they attempt a search with less than 3 characters.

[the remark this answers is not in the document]
dkayiwa: Will the maintainers of this code keep coming to you for an explanation of what kind of warning is represented by this variable?

src/components/bulkConcepts/container/BulkConceptsPage.jsx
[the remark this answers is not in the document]
dkayiwa: Did you ask your team mates?

src/redux/reducers/generalSearchReducer.js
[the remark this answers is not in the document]
brucemakallan: @dkayiwa I'm changing it to MIN_CHARACTERS_WARNING. Is that okay? Would you please suggest something else if it's not?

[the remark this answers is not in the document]
dkayiwa: That looks much better.

src/components/bulkConcepts/container/BulkConceptsPage.jsx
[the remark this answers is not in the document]
brucemakallan: @dkayiwa I was able to get help from the team. I've replaced the 'CIEL' on that line with its constant `INTERNAL_MAPPING_DEFAULT_SOURCE`.

src/redux/reducers/generalSearchReducer.js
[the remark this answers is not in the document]
brucemakallan: @dkayiwa I'm now using `MIN_CHARACTERS_WARNING`

src/components/dashboard/container/DictionariesDisplay.jsx
```diff
@@ -40,10 +42,13 @@ export class DictionaryDisplay extends Component {
...
 
-  onSubmit(event) {
+  onSubmit = (event) => {
     event.preventDefault();
     this.props.clearDictionaries();
-    this.props.searchDictionaries(this.state.searchInput);
+    const { searchInput } = this.state;
+    if (searchInput && searchInput.trim().length > 2) {
+      this.props.searchDictionaries(`${searchInput}*`); // asterisk added to allow partial search
+    } else NotificationManager.warning(MIN_CHARACTERS_WARNING, '', 4000);
```
dkayiwa: What does 4,000 mean?

brucemakallan:
> Does this error message display even when i simply press ENTER?

Yes

src/tests/Dashboard/container/DictionaryDisplay.test.jsx
```diff
@@ -114,17 +114,19 @@ describe('DictionaryDisplay', () => {
...
     const event2 = { target: { name: 'searchInput', value: '', type: 'checkbox' } };
     wrapper.find('#search').simulate('change', event);
     wrapper.find('#search').simulate('change', event2);
-    wrapper.find('#submit-search-form').simulate('submit', {
-      preventDefault: () => {},
+    wrapper.setState({ searchInput: 'test' }, () => {
+      wrapper.update();
+      wrapper.find('#submit-search-form').simulate('submit', { preventDefault: jest.fn() });
+    });
+    wrapper.setState({ searchInput: 'ka' }, () => {
```
dkayiwa: Are you asserting anywhere for this test about expected output?

brucemakallan: @dkayiwa I've fixed this

dkayiwa: Don't you expect specific results?

brucemakallan: The data/results tests are for `Actions` and `Reducers` because they are the ones that make a call to the backend, and change Redux state respectively.

dkayiwa: I thought you would point me to the exact lines of the test you are referring to.

brucemakallan: `dictionaryAction.test.js`

```
it('should return an array of dictionaries on search', () => {
    moxios.wait(() => {
      const request = moxios.requests.mostRecent();
      request.respondWith({
        status: 200,
        response: [dictionaries],
      });
    });

    const expectedActions = [
      { type: IS_FETCHING, payload: true },
      { type: FETCHING_DICTIONARIES, payload: [dictionaries] },
      { type: IS_FETCHING, payload: false },
    ];

    const store = mockStore({ payload: {} });
    return store.dispatch(
      searchDictionaries('', 1000, 1, 'sortAsc=name', 'verbose=true'),
    ).then(() => {
      expect(store.getActions()).toEqual(expectedActions);
    });
  });
```

Partial search is not separate from how search was being done. It just adds an asterisk to ALL search queries so that whether it's a full word or not, it is still searched for.

dkayiwa: If a bug is introduced and results are only returned for full words instead of part of the word, do we have a test which will fail?

brucemakallan: I've added tests which will definitely fail if the asterisk is missing.
```
expect(wrapper.state().searchInput).toEqual('test');
expect(searchDictionaries).toHaveBeenCalledWith(`${wrapper.state().searchInput}*`);
```
Without it, the backend does not return results for partial search.

dkayiwa: Can you make the test expect specific values returned according to the mock data?

brucemakallan: > Can you make the test expect specific values according to the mock data?

That's exactly what the test in Actions (displayed above) is doing. It expects a list of dictionaries to be passed as a payload to the action if that list was returned from the request.

The same test covers both partial search and word search because they work exactly the same way. The only difference would be an asterisk which I have tested for to ensure that it always exists in the search query.

dkayiwa: It expects a list containing how many dictionaries?

brucemakallan: Just one. It's defined as a mock in `dictionaries.js`

dkayiwa: Can you make it expect 3?

brucemakallan: Yes. But the result won't change. What is passed in the `request` section below is what will be expected in the `expectedActions` because it's not actually making a call to the backend. It's just a mock.
```
request.respondWith({
        status: 200,
        response: [dictionaries],
      });
```

Would you still like me to make it 3?

dkayiwa: yes

brucemakallan: I've done so

src/tests/__mocks__/dictionaries.js
```diff
@@ -46,4 +46,22 @@ export const mockDictionaries = (() => {
   return dictionaries;
 })();
 
+export const threeMockDictionaries = [
+  dictionary,
```
dkayiwa: Are these three or two?

src/tests/Dashboard/action/dictionaryAction.test.js
```diff
@@ -38,7 +38,7 @@ import {
   createVersion,
   editMapping,
 } from '../../../redux/actions/dictionaries/dictionaryActionCreators';
-import dictionaries from '../../__mocks__/dictionaries';
+import dictionaries, { threeMockDictionaries } from '../../__mocks__/dictionaries';
```
dkayiwa: The name does not have to include the number of items contained

src/tests/Dashboard/container/DictionaryDisplay.test.jsx
dkayiwa: Can you point me to the line which expects them to be 3?

src/tests/__mocks__/dictionaries.js
brucemakallan: Those are *three*. `dictionary` on line 50 is also a dictionary. The other two are created from that first one with a few changes to avoid code repetition.

src/tests/Dashboard/container/DictionaryDisplay.test.jsx
brucemakallan: There's no such line.
Please note that this Action takes effect AFTER data is received from the backend. It's the backend that performs the actual search. Therefore what we test for is whether what is received is actually passed through the action. It will not have any logic that sifts through the results.

src/tests/Dashboard/action/dictionaryAction.test.js
brucemakallan: Understood. Changing it now.

brucemakallan: I've done so

PR #321, opened 2019-02-01
OCLOMRS-420: fix buffering of add concept on Add CIEL Concepts page

judeinno:
# JIRA TICKET NAME:
[fix buffering of add concept on Add CIEL Concepts page](https://issues.openmrs.org/browse/OCLOMRS-420)

# Summary:
When adding an existing CIEL concept:

- When one clicks the `add concept` button
  *The button is disabled but the concept is not added
- Then when one clicks on another concept to add it.
 *The previous concept is added but the current concept is not added.
- Also when you click the button to add another concept
 *Again the previous concept is added but the current concept is not added.

### DIFF
--- a/src/components/bulkConcepts/component/ActionButtons.jsx
+++ b/src/components/bulkConcepts/component/ActionButtons.jsx
@@ -7,6 +7,8 @@ export class ActionButtons extends Component {
   static propTypes = {
     previewConcept: PropTypes.func.isRequired,
     id: PropTypes.string.isRequired,
+    display_name: PropTypes.string.isRequired,
+    url: PropTypes.string.isRequired,
     params: PropTypes.shape({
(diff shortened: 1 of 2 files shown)


src/components/bulkConcepts/component/ActionButtons.jsx
```diff
@@ -59,6 +60,7 @@ export class ActionButtons extends Component {
 
   render() {
     const { disableButton } = this.state;
+    const url_1 = `/orgs/CIEL/sources/CIEL/concepts/${this.props.id}/`;
```
collinewait: A meaningful name for `url_1` could be better.

judeinno: I have fixed that

PR #326, opened 2019-02-08
OCLOMRS-426: remove autogenerated value from Code field when adding mappings

judeinno:
# JIRA TICKET NAME:
[remove autogenerated value from Code field when adding mappings](https://issues.openmrs.org/browse/OCLOMRS-426)

# Summary:
Remove auto-generated code for to concept code when adding an external mapping
- A user should be able to add an already existing to concept code.

### DIFF
--- a/src/tests/dictionaryConcepts/container/EditConcept.test.jsx
+++ b/src/tests/dictionaryConcepts/container/EditConcept.test.jsx
@@ -267,8 +267,8 @@ describe('Test suite for mappings on existing concepts', () => {
     const event = {
       target: {
         tabIndex: 0,
-        name: 'to_concept_name',
-        value: 'malaria',
+        name: 'to_concept_code',
+        value: 'a234',
       },
     };
     const instance = wrapper.find('EditConcept').instance();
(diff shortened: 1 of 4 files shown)


src/tests/dictionaryConcepts/container/EditConcept.test.jsx
```diff
@@ -274,7 +274,7 @@ describe('Test suite for mappings on existing concepts', () => {
     const instance = wrapper.find('EditConcept').instance();
     instance.updateEventListener({ target: { tabIndex: 0, name: INTERNAL_MAPPING_DEFAULT_SOURCE, value: '' } });
     instance.updateEventListener(event);
-    expect(instance.state.mappings[0].to_concept_code).not.toEqual(null);
```
GodswillOnuoha: You are not to change this. The code is not to be auto-generated, but the field is not to be empty as well. The user is to type of paste the code

judeinno: I have sorted this issue

judeinno: I have sorted this issue

judeinno: Also on the other comment, when a user tries to enter a source that is not the default CIEL, an empty text field is provided with the place holder `To concept code` where the code is supposed to be entered. This is also shown in the video recoding of after the fix.

PR #327, opened 2019-02-08
OCLOMRS-424: Make source field autocomplete, with list of existing sources when adding mapping to concept

brucemakallan:
# JIRA TICKET NAME:
[Make source field autocomplete, with list of existing sources when adding mapping to concept](https://issues.openmrs.org/browse/OCLOMRS-424)

# Summary:
As the user types a source for any given mapping, the source field should automatically display all matching sources (from the existing list of sources).

### DIFF
--- a/src/components/dictionaryConcepts/containers/CreateConcept.jsx
+++ b/src/components/dictionaryConcepts/containers/CreateConcept.jsx
@@ -16,6 +16,8 @@ import {
   addSelectedAnswersToState,
   changeSelectedAnswer,
 } from '../../../redux/actions/concepts/dictionaryConcepts';
+import { fetchConceptSources } from '../../../redux/actions/bulkConcepts';
+import { INTERNAL_MAPPING_DEFAULT_SOURCE, CIEL_SOURCE_URL } from '../components/helperFunction';
 
 export class CreateConcept extends Component {
   static propTypes = {
@@ -52,6 +54,8 @@ export class CreateConcept extends Component {
     addSelectedAnswers: PropTypes.func.isRequired,
     selectedAnswers: PropTypes.array.isRequired,
     changeAnswer: PropTypes.func.isRequired,
+    fetchAllConceptSources: PropTypes.func.isRequired,
+    allSources: PropTypes.arrayOf(PropTypes.shape({})).isRequired,
   };
 
   constructor(props) {
@@ -83,12 +87,13 @@ export class CreateConcept extends Component {
       match: {
         params: { conceptType },
       },
+      fetchAllConceptSources,
     } = this.props;
     const concept = conceptType || '';
     this.props.createNewName();
     this.props.addNewDescription();
-    // eslint-disable-next-line
     this.setState({ concept_class: concept });
+    fetchAllConceptSources();
   }
 
   componentWillReceiveProps(nextProps) {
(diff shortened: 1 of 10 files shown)


brucemakallan:
screen shot 2019-02-08 at 15 53 53

Currently waiting on Ticket OCLOMRS-406 which adds edit functionality for mappings on the right interface so I can add the autocomplete feature to it as well.

src/tests/dictionaryConcepts/container/EditConcept.test.jsx
```diff
@@ -277,6 +284,18 @@ describe('Test suite for mappings on existing concepts', () => {
...
 
+  it('should call updateEventListener function with internal mapping', () => {
+    const event = {
+      target: {
+        tabIndex: 0,
+        name: 'to_concept_name',
+        value: INTERNAL_MAPPING_DEFAULT_SOURCE,
+      },
+    };
+    const instance = wrapper.find('EditConcept').instance();
```
dkayiwa: Shouldn't we expect something here?

src/tests/dictionaryConcepts/container/CreateConcept.test.jsx
```diff
@@ -93,6 +96,48 @@ describe('Test suite for dictionary concepts components', () => {
...
 
+  it('should call updateEventListener function with internal mapping', () => {
+    const event = {
+      target: {
+        tabIndex: 0,
+        name: 'to_concept_name',
+        value: INTERNAL_MAPPING_DEFAULT_SOURCE,
+      },
+    };
+    const instance = wrapper.find('CreateConcept').instance();
```
dkayiwa: Shouldn't we expect something here?

src/tests/dictionaryConcepts/container/EditConcept.test.jsx
brucemakallan: I have corrected this.

src/tests/dictionaryConcepts/container/CreateConcept.test.jsx
brucemakallan: I have corrected this as well. Sorry about that.

src/components/dictionaryConcepts/containers/CreateConcept.jsx
```diff
@@ -253,10 +258,30 @@ export class CreateConcept extends Component {
   updateEventListener = (event) => {
     const { tabIndex, name, value } = event.target;
     const { mappings } = this.state;
-    mappings[tabIndex][name] = value;
+    // if the ciel source URL was provided, use just the name
+    if (value.toUpperCase().match(INTERNAL_MAPPING_DEFAULT_SOURCE)) {
```
dkayiwa: Why just use the name in this case?

brucemakallan: @dkayiwa When the user types a source and selects from the dropdown, what the app takes is the URL to that source and not the displayed name (because the backend takes a URL).
So what I'm doing here is checking to see if that value (URL) contains CIEL (a special case) so that I can save the CIEL as it is and not its URL.

dkayiwa: Contains CIEL or is equal to CIEL?

brucemakallan: If the URL of the source `contains` CIEL. All other sources will not contain CIEL

dkayiwa: So what about if the source contains something like CIELBRUCE?

brucemakallan: It will take the value CIEL. Let me filter by the entire URL to prevent that from happening.

brucemakallan: @dkayiwa I have made the correction

dkayiwa: Have you tested the application again with this change to confirm that it still works?

brucemakallan: Yes. I manually tested both `Create Concept` and `Edit Concept` with the concept name `HS1` in the `Kampala Dictionary`

PR #329, opened 2019-02-11
OCLOMRS-404: make source name case insensitive on mappings

judeinno:
# JIRA TICKET NAME:
[make source name case insensitive on mappings](https://issues.openmrs.org/browse/OCLOMRS-404)

# Summary:
When a user enters a source, the text field:

- Should be case insensitive, that is, allows both upper and lower case

### DIFF
--- a/src/components/dictionaryConcepts/components/CreateMapping.jsx
+++ b/src/components/dictionaryConcepts/components/CreateMapping.jsx
@@ -16,6 +16,8 @@ class CreateMapping extends Component {
     this.setState({ inputValue: value });
   }
 
+  sourceNameToUpperCase = source => source.trim().toUpperCase();
+
   render() {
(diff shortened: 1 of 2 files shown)


src/components/dictionaryConcepts/components/CreateMapping.jsx
```diff
@@ -16,6 +16,8 @@ class CreateMapping extends Component {
     this.setState({ inputValue: value });
   }
 
+  sourceCaseSensitivity = source => source.trim().toUpperCase();
```
GodswillOnuoha: I think `sourceCaseSensitivity` can be given a name that better tells what it does.

judeinno: I have changed it to a better name `sourceNameToUpperCase`

PR #339, opened 2019-02-21
OCLOMRS-451: The remove button on the name and description column of creating new concept does not work

judeinno:
# JIRA TICKET NAME:
[The remove button on the name and description column of creating new concept does not work](https://issues.openmrs.org/browse/OCLOMRS-451)

# Summary:
On the create new concept page:
- The button for remove name does not function as expected. 
- Also the remove description button, not function as expected.

### DIFF
--- a/src/components/dictionaryConcepts/components/DescriptionRow.jsx
+++ b/src/components/dictionaryConcepts/components/DescriptionRow.jsx
@@ -2,30 +2,32 @@ import React, { Component } from 'react';
 import Select from 'react-select';
 import autoBind from 'react-autobind';
 import PropTypes from 'prop-types';
-import uuid from 'uuid/v4';
 import locale from '../../dashboard/components/dictionary/common/Languages';
 
 class DescriptionRow extends Component {
   static propTypes = {
-    newRow: PropTypes.shape({
-      uuid: PropTypes.string,
-      locale: PropTypes.string,
-      description: PropTypes.string,
-    }),
+    newRow: PropTypes.object,
+    newRowUuid: PropTypes.string,
     addDataFromDescription: PropTypes.func.isRequired,
     removeDescription: PropTypes.func.isRequired,
     removeDataFromRow: PropTypes.func.isRequired,
     pathName: PropTypes.object.isRequired,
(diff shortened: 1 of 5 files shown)


AGMETEOR:
but fix the drop in coverage...

dkayiwa:
Any reason why this is still open for a full week?

judeinno:
Yes, you had requested for the remove labels to be changed to buttons
before you can merge it. And the PR is up for changing the buttons just
waiting for all 5 approval then request for merging.
After that then I can request for merging this one again.

src/components/dictionaryConcepts/components/DescriptionRow.jsx
```diff
@@ -124,7 +126,9 @@ class DescriptionRow extends Component {
             className=" btn btn-danger concept-form-table-link"
             id="remove-description"
             type="button"
-            onClick={event => this.handleRemove(event, this.props.newRow)}
+            onClick={event => ((this.props.newRowUuid === '')
```
dkayiwa: Why not instead check for null?

judeinno: Because the value am checking against is a string and null is of object type

dkayiwa: What is the difference between newRow and newRowUuid?

judeinno: newRow is used on the Edit concept interface to access a specific description row details and newRowUuid is used when on the create concepts page when accessing a specific description row id that is stored in a list called description.

judeinno: The Implementaion for the adding of a description row on the edit concept page and the create concept page are different. This is where the two different terms arise from.
When adding a new decription row on the edit concept page.
- This stores the row data in state as an object with key as Uuid so when removing the row we access the id as newRow.uuid

while
When adding a new decription row on the create concept page.
- This stores the row data in state as a list of UUIDs in form of strings so when removing the row we access the id directly

Hence the two different term

PR #346, opened 2019-02-27
OCLOMRS-445: Fix pagination on Add CIEL concepts page when datatype/class filters are selected.

AGMETEOR:
# JIRA TICKET NAME:
[Fix pagination on Add CIEL concepts page when datatype/class filters are selected.](https://issues.openmrs.org/browse/OCLOMRS-445)

# Summary:
On the Add CIEL concepts page;
- Once a user selects concept by datatype or class.
 * The pagination buttons at the right bottom corner of the table stop functioning.
 * That is they change the pagination number but the concepts being displayed do not change.

 * Fix the pagination by ensuring that a user can select a  specific datatype/class and view the next and previous page when they click the pagination buttons.

### DIFF
--- a/src/tests/bulkConcepts/containers/BulkConceptsPage.test.js
+++ b/src/tests/bulkConcepts/containers/BulkConceptsPage.test.js
@@ -8,6 +8,7 @@ import {
   mapStateToProps,
 } from '../../../components/bulkConcepts/container/BulkConceptsPage';
 import concepts from '../../__mocks__/concepts';
+import { INTERNAL_MAPPING_DEFAULT_SOURCE } from '../../../components/dictionaryConcepts/components/helperFunction';
 
 const store = createMockStore({
   bulkConcepts: {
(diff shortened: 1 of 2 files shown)


src/tests/bulkConcepts/containers/BulkConceptsPage.test.js
```diff
@@ -333,6 +334,38 @@ describe('Test suite for BulkConceptsPage component', () => {
...
+      addToFilterList: jest.fn(),
+      fetchFilteredConcepts: jest.fn(),
+      addConcept: jest.fn(),
+      previewConcept: jest.fn(),
+      handleNextPage: jest.fn(),
+      componentDidUpdate: jest.fn(),
+    };
+    const wrapper = shallow(<BulkConceptsPage {...props} />).setState({ filterOn: true });
+    const instance = wrapper.instance();
+    expect(instance.componentDidUpdate(2)).toEqual(undefined);
```
dkayiwa: How is this test related to the pagination fix?

AGMETEOR: @dkayiwa , yes, it's related.

dkayiwa: How?

AGMETEOR: The test is similar to the one just above it for a normal search and then pagination only that for this particular one, filtering in true in the state.

dkayiwa: So does this test fail when i set filtering to false?

AGMETEOR: No, it doesn't. Let me modify the test or take it out.

dkayiwa: Then it is not testing what it should.

AGMETEOR: @dkayiwa I think I made a mistake on using the test just before this one as a template. I have now improved this particular test.

src/tests/bulkConcepts/containers/BulkConceptsPage.test.js
```diff
@@ -333,6 +334,39 @@ describe('Test suite for BulkConceptsPage component', () => {
...
+      addToFilterList: jest.fn(),
+      fetchFilteredConcepts: jest.fn(),
+      addConcept: jest.fn(),
+      previewConcept: jest.fn(),
+      handleNextPage: jest.fn(),
+    };
+    const wrapper = shallow(<BulkConceptsPage {...props} />);
+    const instance = wrapper.instance();
+    wrapper.setState({ filterOn: true });
+    instance.componentDidUpdate(props.currentPage + 1);
```
dkayiwa: Will this test fail if the above page is not incremented?

AGMETEOR: No, it will not. What the test actually looks for is whether an update is made in the component and then uses the current page to fetch the concepts. Let me change the description so that it's not misleading in terms of what the test does.

dkayiwa: https://en.wikipedia.org/wiki/Test-driven_development

AGMETEOR: I updated the test to remove the increment in the current page props.

PR #350, opened 2019-03-06
OCLOMRS-453: Fix inability to create more than one Name or Description while Creating and Editing Concepts

brucemakallan:
# JIRA TICKET NAME:
[Fix inability to create more than one Name or Description while Creating and Editing Concepts](https://issues.openmrs.org/browse/OCLOMRS-453)

# Summary:
This bug prevents the creation of more than one name and more than one description while creating or editing a custom concept. The concept itself is saved successfully, but the new name or new description is not saved.

This happens because:

- The identifiers for each new name-row and description-row were not passed correctly.
- The default locale (a required field) wasn't being passed correctly.

### DIFF
--- a/src/components/dictionaryConcepts/components/ConceptNameRows.jsx
+++ b/src/components/dictionaryConcepts/components/ConceptNameRows.jsx
@@ -14,6 +14,7 @@ class ConceptNameRows extends Component {
     removeDataFromRow: PropTypes.func.isRequired,
     pathName: PropTypes.object.isRequired,
     existingConcept: PropTypes.object.isRequired,
+    rowId: PropTypes.string,
   };
 
   static defaultProps = {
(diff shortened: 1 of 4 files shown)


src/components/dictionaryConcepts/components/ConceptNameRows.jsx
```diff
@@ -169,7 +164,7 @@ class ConceptNameRows extends Component {
             className="btn btn-danger concept-form-table-link"
             id="remove-name"
             type="button"
-            onClick={event => this.handleRemove(event, this.props.newRow.uuid)}
+            onClick={event => this.handleRemove(event, this.props.rowId)}
```
Mrshinezee: I think you should always destructure ur props

brucemakallan: Thanks. I have made the correction.

PR #354, opened 2019-03-11
OCLOMRS-461: The source of the Mapping does not display while editing a Concept

brucemakallan:
# JIRA TICKET NAME:
[The source of the Mapping does not display while editing a Concept](https://issues.openmrs.org/browse/OCLOMRS-461)

# Summary:
When a concept is created with its mappings and then the user clicks to edit the concept, the mappings display without their sources.

### DIFF
--- a/src/components/dictionaryConcepts/components/CreateMapping.jsx
+++ b/src/components/dictionaryConcepts/components/CreateMapping.jsx
@@ -31,7 +31,7 @@ class CreateMapping extends Component {
             className="form-control"
             name="source"
             onChange={(event) => { updateEventListener(event, url); }}
-            defaultValue={isNew ? '' : source}
+            value={source || undefined}
           >
             <option value="" hidden>Select a source</option>
             {allSources.map(src => <option


src/components/dictionaryConcepts/components/CreateMapping.jsx
```diff
@@ -31,7 +31,7 @@ class CreateMapping extends Component {
             className="form-control"
             name="source"
             onChange={(event) => { updateEventListener(event, url); }}
-            defaultValue={isNew ? '' : source}
+            value={source || undefined}
           >
             <option value="" hidden>Select a source</option>
```
brucemakallan: When the value property is set to an empty string, it will select the option that says `"Select a source"` because their values will match. However, if it is set to `undefined` it will adopt the value selected by the user from the dropdown (when Creating) and the correct source value (when editing).

Mrshinezee: Ok

PR #357, opened 2019-03-13
OCLOMRS-463: Passing OCL URL environment variables to run at runtime

judeinno:
# JIRA TICKET NAME:
[Passing OCL URL environment variables to run at runtime](https://issues.openmrs.org/browse/OCLOMRS-463)

# Summary:
- Create a bash script that creates an env-config.js file with key-value pairs of the environment variables.
- Load env-config.js  with the script tag in index.html
- Update the start script in package.json to run the bash script when running the app during development.
- Update the Dockerfile to copy the .env file, the bash script and also run the bash script before the build(at runtime). Then remove env-config.js from the public folder if any and add the newly created one by the bash script. 
 

**Testing.**

**In development**

- Create a .env file in the root directory and pass it;

    
`OCL_API_HOST=https://api.qa.openconceptlab.org/`
`TRADITIONAL_OCL_HOST=https://qa.openconceptlab.org`

- Then run `npm start

**In docker**

- ** Create a .env file in the root directory and pass it;

`OCL_API_HOST=https://api.qa.openconceptlab.org/`
`TRADITIONAL_OCL_HOST=https://qa.openconceptlab.org`

- Then run `docker build -t openmrs/ocl-client:local .`
- Wait for the build to finish, then run `docker-compose -f docker/docker-compose.yml up ` to run the Docker container
- Wait for the container to run, then access the app at `http://localhost:8081/`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -39,5 +39,16 @@ COPY --from=build-deps /usr/src/app/docker/default.conf /etc/nginx/conf.d/
 # Make port 80 available to the world outside the container
 EXPOSE 80
 
-# Run the nginx server
-CMD ["nginx", "-g", "daemon off;"]
+# Copy .env file and shell script to container
+WORKDIR /usr/share/nginx/html
+COPY ./init.sh .
+COPY .env .
+
+# Add bash
+RUN apk add --no-cache bash
+
+# Make our shell script executable
+RUN chmod +x init.sh
+
+# Start Nginx server
+CMD ["/bin/bash", "-c", "/usr/share/nginx/html/init.sh && nginx -g \"daemon off;\""]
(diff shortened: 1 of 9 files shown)


Dockerfile
```diff
@@ -23,6 +23,22 @@ RUN npm install react-scripts@1.1.1 -g --silent
...
 
+# # Copy .env file and shell script to container
+COPY ./init.sh .
+COPY .env .
+
+# Make our shell script executable
+RUN chmod +x init.sh
+
+# Run the bash script
+RUN /bin/bash -c ./init.sh
```
rkorytkowski: The whole point of having the bash script is to run it at runtime in the CMD docker command (see the very bottom) and not during the build. We want to change environment properties AT RUNTIME using a pre-built image.

judeinno: Let me work on that

judeinno: I have fixed the issue please review.

PR #359, opened 2019-03-15
OCLOMRS-462: When a concept mapping from a specific source is removed  a user should be able to re-add it

alexkayabula:
# JIRA TICKET NAME:
[When a concept mapping from a specific source is removed  a user should be able to add it back ](https://issues.openmrs.org/browse/OCLOMRS-462)

# Summary:
When editing a concept, a user should be able to remove a concept mapping and re-add it. When one removes an already existing mapping then selects the same concept mapping and clicks update, the concept mapping is not added.

### DIFF
--- a/src/redux/actions/concepts/dictionaryConcepts.js
+++ b/src/redux/actions/concepts/dictionaryConcepts.js
@@ -369,3 +369,15 @@ export const updateConcept = (conceptUrl, data, history) => async (dispatch) => 
 export const clearPreviousConcept = () => (dispatch) => {
   dispatch({ type: CLEAR_PREVIOUS_CONCEPT });
 };
+
+export const unretireMapping = url => async (dispatch) => {
+  dispatch(isFetching(true));
+  try {
+    const response = await instance.put(url, { retired: false });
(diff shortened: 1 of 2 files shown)


src/redux/actions/concepts/dictionaryConcepts.js
```diff
@@ -369,3 +369,15 @@ export const updateConcept = (conceptUrl, data, history) => async (dispatch) =>
 export const clearPreviousConcept = () => (dispatch) => {
   dispatch({ type: CLEAR_PREVIOUS_CONCEPT });
 };
+
+export const unretireMapping = url => async (dispatch) => {
+  dispatch(isFetching(true));
+  try {
+    const response = await instance.put(url, { retired: false });
```
dkayiwa: How is this related to the ticket?

alexkayabula: When a concept mapping is removed, the concept is actually retired. For that specific concept mapping to be re-added, it has to be unretired i.e the retired flag has to be updated to false.

PR #360, opened 2019-03-15
OCLOMRS-470: Make “Narrower-Than” the default relationship for external mappings, and “SAME-AS” the default for CIEL (when adding a mapping)

judeinno:
# JIRA TICKET NAME:
[Make “Narrower-Than” the default relationship for external mappings, and “SAME-AS” the default for CIEL (when adding a mapping)](https://issues.openmrs.org/browse/OCLOMRS-470)

# Summary:
Make “Narrower-Than” the default relationship for external mappings (all mappings except CIEL) when adding or editing a concept mapping.

### DIFF
--- a/src/components/dictionaryConcepts/components/MapType.jsx
+++ b/src/components/dictionaryConcepts/components/MapType.jsx
@@ -1,14 +1,15 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import { INTERNAL_MAPPING_DEFAULT_SOURCE, MAP_TYPES_DEFAULTS } from './helperFunction';
 
 const mapType = (props) => {
   const {
-    index, map_type, map_types, updateEventListener, url,
+    index, map_type, map_types, updateEventListener, url, source,
   } = props;
   return (
     <select
       tabIndex={index}
-      defaultValue={map_type}
+      value={source !== INTERNAL_MAPPING_DEFAULT_SOURCE ? MAP_TYPES_DEFAULTS[1] : map_type}
       className="form-control"
       placeholder="map type"
       type="text"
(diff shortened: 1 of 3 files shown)


src/components/dictionaryConcepts/components/MapType.jsx
```diff
@@ -3,12 +3,12 @@ import PropTypes from 'prop-types';
...
 const mapType = (props) => {
   const {
-    index, map_type, map_types, updateEventListener, url,
+    index, map_type, map_types, updateEventListener, url, source,
   } = props;
   return (
     <select
       tabIndex={index}
-      defaultValue={map_type}
+      value={source !== 'CIEL' ? 'NARROWER-THAN' : map_type}
```
dkayiwa: Shouldn't this be already in a constant?

judeinno: I have fixed it and added the constant `INTERNAL_MAPPING_DEFAULT_SOURCE` used for `'CIEL'`

src/components/dictionaryConcepts/components/MapType.jsx
```diff
@@ -33,12 +34,14 @@ mapType.propTypes = {
...
   map_types: PropTypes.array,
   url: PropTypes.string,
+  source: PropTypes.string,
 };
 
 mapType.defaultProps = {
   url: '',
   map_type: '',
   index: 0,
+  source: 'CIEL',
```
dkayiwa: Don't we already have a constant for that?

src/components/dictionaryConcepts/components/MapType.jsx
```diff
@@ -1,14 +1,15 @@
...
 const mapType = (props) => {
   const {
-    index, map_type, map_types, updateEventListener, url,
+    index, map_type, map_types, updateEventListener, url, source,
   } = props;
   return (
     <select
       tabIndex={index}
-      defaultValue={map_type}
+      value={source !== INTERNAL_MAPPING_DEFAULT_SOURCE ? 'NARROWER-THAN' : map_type}
```
dkayiwa: Shouldn't we use a constant for the above?

src/components/dictionaryConcepts/components/MapType.jsx
judeinno: I have fixed it.

src/tests/Signup/index.test.jsx
```diff
@@ -28,4 +28,60 @@ describe('Sign up Component', () => {
     wrapper.find('#firstName').simulate('change', event);
     expect(spy).toHaveBeenCalledTimes(1);
   });
+
+  it('should handle text lastname input', () => {
+    const event = {
```
dkayiwa: How is this test related to the ticket?

judeinno: Test coverage had dropped and when I checked where the drop was coming from it showed that the drop was coming from it showed the drop was arising from Sign up as shown in the screenshot below so I worked on raising the coverage
image

PR #373, opened 2019-03-22
OCLOMRS-484: Display mappings on the Q-AND-A pages both on the Create and Edit page

AGMETEOR:
# JIRA TICKET NAME:
[Display mappings on the Q-AND-A pages both on the Create and Edit page](https://issues.openmrs.org/browse/OCLOMRS-484)

# Summary:
Currently, on the page where we either create or edit Q-AND-A concepts, there are no mappings shown. In this ticket, on the page where we create Q-AND-A concepts, fields for adding mappings should be displayed at the bottom of the page. Likewise, on the page where we edit a Q-AND-A concept, mappings should be displayed at the bottom of the page.

### DIFF
--- a/src/components/dictionaryConcepts/components/CreateConceptForm.jsx
+++ b/src/components/dictionaryConcepts/components/CreateConceptForm.jsx
@@ -5,7 +5,7 @@ import CreateConceptTable from './CreateConceptTable';
 import DescriptionTable from './DescriptionTable';
 import AnswersTable from './AnswersTable';
 import CreateMapping from './CreateMapping';
-import { classes } from './helperFunction';
+import { classes, MAP_TYPE } from './helperFunction';
 
 const CreateConceptForm = (props) => {
   const {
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -30,7 +30,7 @@
     "uuid": "^3.3.2"
   },
   "scripts": {
-    "start": "chmod +x ./init.sh && ./init.sh && cp env-config.js ./public/ && npm run build:css && react-scripts start",
```
dkayiwa: How is this related to the ticket?

AGMETEOR: @dkayiwa that script stops the application from running and removing it is the only way the application will start.

dkayiwa: Was that caused by changes in this ticket? If not, then it should be a separate ticket and pull request.

AGMETEOR: @dkayiwa no it was not. There's a ticket that's been put up to solve the issue with the start script. Let me rebase.

dkayiwa:
Did you just forget to add tests for these changes?

AGMETEOR:
@dkayiwa when I look at the coverage file, the current tests cover the lines added.

PR #384, opened 2019-04-03
OCLOMRS-490: Fix and add remove button on dictionary concepts page with class Concept Class

danuluma:
# JIRA TICKET NAME:
[OCLOMRS-490](https://issues.openmrs.org/browse/OCLOMRS-490)

# Summary:
While viewing Concepts of a specific Dictionary, the "Remove" button/action is also be available for Concepts with the class, "Concept Class"


danuluma:
This PR is being closed because the bug was on concepts added via the “Other” option on the “Bulk Add Concepts” page, and this option has since been removed.

PR #390, opened 2019-04-04
OCLOMRS-498: Remove the “Other” Option from Bulk Add Concepts page

judeinno:
# JIRA TICKET NAME:
[Remove the “Other” Option from Bulk Add Concepts page](https://issues.openmrs.org/browse/OCLOMRS-498)

# Summary:
Remove the Other radio button on the Bulk Add Concepts page it will be re-implemented in post MVP.

### DIFF
--- a/src/components/bulkConcepts/addBulkConcepts.jsx
+++ b/src/components/bulkConcepts/addBulkConcepts.jsx
@@ -40,7 +40,6 @@ export class AddBulkConcepts extends Component {
     this.state = {
       conceptIds: '',
       openResultModal: false,
-      otherSelected: false,
       value: '',
     };
     this.invalidConceptIds = [];
@@ -58,22 +57,10 @@ export class AddBulkConcepts extends Component {
   }
 
   handleCielClick = (e) => {
-    this.setState({ otherSelected: false });
     this.sourceUrl = e.target.value;
     this.props.fetchSourceConcepts(this.sourceUrl);
     this.reset();
   };
-
-  otherSourceClick = () => {
-    this.setState({ otherSelected: true });
-    this.props.fetchConceptSources();
-    this.reset();
-  }
-
-  handleSourceSelect = (source) => {
-    this.sourceUrl = source.url.substring(1);
-    this.props.fetchSourceConcepts(this.sourceUrl);
-  }
 
   reset = () => {
     this.setState({ conceptIds: '' });
@@ -127,7 +114,7 @@ export class AddBulkConcepts extends Component {
 
   render() {
     const dictionaryName = localStorage.getItem('dictionaryName');
-    const { conceptIds, openResultModal, otherSelected } = this.state;
+    const { conceptIds, openResultModal } = this.state;
     const {
       match: {
         params: {
@@ -135,9 +122,7 @@ export class AddBulkConcepts extends Component {
         },
       },
       sourceConcepts,
-      conceptSources,
       isLoading,
-      isFetching,
     } = this.props;
     const disableButton = (conceptIds.length < 1);
     return (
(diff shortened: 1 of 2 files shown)


src/components/bulkConcepts/addBulkConcepts.jsx
```diff
@@ -58,23 +57,11 @@ export class AddBulkConcepts extends Component {
...
   };
 
-  otherSourceClick = () => {
-    this.setState({ otherSelected: true });
-    this.props.fetchConceptSources();
-    this.reset();
-  }
-
-  handleSourceSelect = (source) => {
-    this.sourceUrl = source.url.substring(1);
```
dkayiwa: How is this related to removing other?

src/components/bulkConcepts/addBulkConcepts.jsx
```diff
@@ -127,17 +114,15 @@ export class AddBulkConcepts extends Component {
...
-    const { conceptIds, openResultModal, otherSelected } = this.state;
+    const { conceptIds, openResultModal } = this.state;
     const {
       match: {
         params: {
           type, typeName, collectionName, language,
         },
       },
       sourceConcepts,
-      conceptSources,
```
dkayiwa: How is this related to removing other?

src/components/bulkConcepts/addBulkConcepts.jsx
judeinno: This was being used by the other search to search for the sources.

src/components/bulkConcepts/addBulkConcepts.jsx
judeinno: This was used by the others search field to filter the source names by the input data provided. This is shown on line 228 that I had to clear.

src/tests/bulkConcepts/addBulkConcepts.test.jsx
```diff
@@ -74,27 +74,6 @@ describe('Add Bulk Concepts', () => {
     expect(wrapper.find('Loader')).toHaveLength(1);
   });
 
-  it('should render a loader while fetching Concepts Source', () => {
-    const properties = {
-      fetchSourceConcepts: jest.fn(),
-      addExistingBulkConcepts: jest.fn(),
-      fetchConceptSources: jest.fn(),
```
dkayiwa: What happens if we leave this test?

src/tests/bulkConcepts/addBulkConcepts.test.jsx
```diff
@@ -207,39 +186,6 @@ describe('Add Bulk Concepts', () => {
...
-    const wrapper = mount(<MemoryRouter>
-      <Provider store={store}>
-        <AddBulkConcepts {...props} />
-      </Provider>
-    </MemoryRouter>);
-    const other = wrapper.find('#otherSourcesOption');
-    other.simulate('click');
-    const input = wrapper.find('#sourceSearch');
-    input.simulate('change', { target: { value: 'testSource' } });
-    input.simulate('keydown', { key: 'ArrowDown' });
```
dkayiwa: What happens if we leave this test?

judeinno: It will fail because its searching through the component to find an Id/node `#otherSourcesOption`that does not exist hence it will throw an error saying. `This method is only meant to be run on single node. 0 found instead.`

src/tests/bulkConcepts/addBulkConcepts.test.jsx
judeinno: It will fail because it will try to find a loader of length 1 but instead find length 0 because since we are searching from non-CIEL concepts it will search through the other sources and since that field no longer exists the test will fail.

PR #392, opened 2019-04-05
OCLOMRS-499: Fix memory leak console log error when running tests

alexkayabula:
# JIRA TICKET NAME:
[Fix memory leak console log error when running tests](https://issues.openmrs.org/browse/OCLOMRS-499)

# Summary:
When running tests, a console log error associated with createMappings.test appears that indicates a memory leak in the application. This could result in performance issues.

### DIFF
--- a/src/components/dictionaryConcepts/components/CreateMapping.jsx
+++ b/src/components/dictionaryConcepts/components/CreateMapping.jsx
@@ -47,9 +47,11 @@ class CreateMapping extends Component {
     }
   }
 
-  handleKeyPress = (event, inputValue, url) => {
+
+  handleKeyPress = async (event, inputValue, url) => {
     if (event.keyCode === KEY_CODE_FOR_ENTER) {
-      this.doSearch(inputValue, url);
+      const options = await fetchSourceConcepts(INTERNAL_MAPPING_DEFAULT_SOURCE, inputValue, url);
+      this.setState({ options, isVisible: true });
     }
   }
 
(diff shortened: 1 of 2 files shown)


src/components/dictionaryConcepts/components/CreateMapping.jsx
```diff
@@ -47,7 +56,9 @@ class CreateMapping extends Component {
     }
   }
 
+
   handleKeyPress = (event, inputValue, url) => {
+    // eslint-disable-next-line no-underscore-dangle
```
paulvazbad: Hey! I think it would be greatt to remove the eslint comments from the file

src/components/dictionaryConcepts/components/CreateMapping.jsx
```diff
@@ -60,7 +71,10 @@ class CreateMapping extends Component {
 
   doSearch = async (inputValue, url) => {
     const options = await fetchSourceConcepts(INTERNAL_MAPPING_DEFAULT_SOURCE, inputValue, url);
-    this.setState({ options, isVisible: true });
+    // eslint-disable-next-line no-underscore-dangle
+    if (this._isMounted) {
+      this.setState({ options, isVisible: true });
```
paulvazbad: isVisible should have an inicial value in the state definition of the object.

src/components/dictionaryConcepts/components/CreateMapping.jsx
alexkayabula: It has been fixed.

src/components/dictionaryConcepts/components/CreateMapping.jsx
alexkayabula: It already has an initial value in the state definition.

PR #394, opened 2019-04-08
OCLOMRS-501: Refreshing the Create Concept page should not render new rows for Names and Descriptions

brucemakallan:
# JIRA TICKET NAME:
[Refreshing the Create Concept page should not render new rows for Names and Descriptions](https://issues.openmrs.org/browse/OCLOMRS-501)

# Summary:
When the Create Concept page is refreshed, new rows for Names and Descriptions are created.

### DIFF
--- a/src/tests/dictionaryConcepts/container/CreateConcept.test.jsx
+++ b/src/tests/dictionaryConcepts/container/CreateConcept.test.jsx
@@ -134,6 +134,28 @@ describe('Test suite for dictionary concepts components', () => {
     expect(wrapper).toMatchSnapshot();
   });
 
+  it('should not call the createNewName function if new names are already available', () => {
+    const propsWithNames = { ...props, newName: ['a'] };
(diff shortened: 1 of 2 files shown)


dkayiwa:
The new tests that you have added are good. But i still do not see tests to address the bug that was fixed.

src/tests/dictionaryConcepts/container/CreateConcept.test.jsx
```diff
@@ -134,6 +134,28 @@ describe('Test suite for dictionary concepts components', () => {
     expect(wrapper).toMatchSnapshot();
   });
 
+  it('should not call the createNewName function if new names are already available', () => {
```
brucemakallan: @dkayiwa The bug was there because `createNewName()` was being called when a new name was already available. I have therefore **added a test** to ensure that when a new name is present, the function is not called. i.e. **not**.toHaveBeenCalled()

PR #400, opened 2019-04-16
OCLOMRS-507: source does not show all available sources	

myrdstom:
# JIRA TICKET NAME:
[when creating a mapping, source dropdown does not show all available sources](https://issues.openmrs.org/browse/OCLOMRS-507)

# Summary:
source does not show all available sources

### DIFF
--- a/src/redux/actions/bulkConcepts/index.js
+++ b/src/redux/actions/bulkConcepts/index.js
@@ -74,7 +74,7 @@ export const fetchConceptSources = () => async (dispatch) => {
   dispatch(isFetching(true));
   dispatch(clear(CLEAR_SOURCE_CONCEPTS));
   try {
-    const response = await instance.get('/sources/');
+    const response = await instance.get('/sources/?limit=0');
     dispatch(isSuccess(response.data, FETCH_CONCEPT_SOURCES));
     dispatch(isFetching(false));
   } catch (error) {


src/redux/actions/bulkConcepts/index.js
```diff
@@ -74,7 +74,7 @@ export const fetchConceptSources = () => async (dispatch) => {
   dispatch(isFetching(true));
   dispatch(clear(CLEAR_SOURCE_CONCEPTS));
   try {
-    const response = await instance.get('/sources/');
+    const response = await instance.get('/sources/?limit=0');
```
dkayiwa: What exactly does this change do?

myrdstom: By default, the endpoint is paginated to display only the first 10 results for a source. The result of this is the user has access to only the first 10 `sources`. 
 What `?limit=0` does is it removes the pagination limit

dkayiwa: And this? https://github.com/myrdstom/openmrs-ocl-client/commit/895dad284c81e0b025fb7ae91a120b0d187b5f1a

myrdstom: I have responded to it. I must have made a mistake while pushing code and reverted

dkayiwa: So what happens when the sources are too many? Isn't this the reason why we had paging in the first place?

myrdstom: If the sources are many, there is a scrollable dropdown menu that appears

dkayiwa: Scrolling through a million sources loaded at once?

myrdstom: Currently, we have 41 sources on the demo server which are currently manageable with a dropdown. The only other option would be to make it a search and it could imply that the relationship mapping would also have to be made a search. 

Given that we would be making major changes to the application, I would think it wise to pass this through the product owners during the call today and get their opinion of the matter, what do you think?

PR #401, opened 2019-04-17
OCLOMRS-508: Fix the position and make the drop down search menu on the Add Bulk concepts page scrollable

alexkayabula:
# JIRA TICKET NAME:
[Fix the position and make the drop down search menu on the Add Bulk concepts page scrollable](https://issues.openmrs.org/browse/OCLOMRS-508)

# Summary:
When adding bulk concepts, searching for CIEL concepts might return a drop down of results in hundreds or thousands. This could make the display uncharacteristically inconveniencing and bad for the users. The results drop down should be made scrollable to improve user experience.


src/styles/bulk_concepts.scss
```diff
@@ -215,12 +215,18 @@
   align-items: center;
 }
 
-.search-ul{
```
dkayiwa: What is the reason behind renaming it?

alexkayabula: Previously the <div> tag had the same className as the <ul> tag on which I had to make the changes in CSS.
In order to apply the changes specifically to the <ul> tag, I had to change its className.

PR #425, opened 2019-04-30
WIP-OCLOMRS-524: When Editing a dictionary the added languages don't show up after updating

alexkayabula:
# JIRA TICKET NAME:
[When Editing a dictionary the added languages don't show up after updating](https://issues.openmrs.org/browse/OCLOMRS-524)

# Summary:
When editing a dictionary and adding some other languages, when the dictionary is updated and Edit clicked again, the popup does not show up with the additional languages. After refreshing the page, the languages do show up.

### DIFF
--- a/src/components/dashboard/components/dictionary/DictionaryContainer.jsx
+++ b/src/components/dashboard/components/dictionary/DictionaryContainer.jsx
@@ -78,7 +78,9 @@ export class DictionaryOverview extends Component {
     }
   }
 
-  handleHide = () => this.setState({ showEditModal: false });
+  handleHide = () => {
+    this.setState({ showEditModal: false });
+  };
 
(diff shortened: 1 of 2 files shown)


dkayiwa:
Shouldn't you have a test that fails without this change?

alexkayabula:
> Shouldn't you have a test that fails without this change?

No there is not, because what was actually happening was the languages were being added but since the dictionaryModal was not re-rendered after submitting the updated changes, the added languages could not show up.

dkayiwa:
My point is, shouldn't you add a test to reproduce the reported problem?

alexkayabula:
> My point is, shouldn't you add a test to reproduce the reported problem?

Let me add the test.

PR #429, opened 2019-05-02
OCLOMRS-552: Enable OpenMRS validation for all dictionaries created to prevent duplication

alexkayabula:
# JIRA TICKET NAME:
[Enable OpenMRS validation for all dictionaries created to prevent duplication](https://issues.openmrs.org/browse/OCLOMRS-552)

# Summary:
Enable OpenMRS validation on the dictionaries we create through this application. This will prevent the dictionary from having two concepts with identical preferred names, which is illegal in OpenMRS.

### DIFF
--- a/src/components/dashboard/components/dictionary/common/DictionaryModal.jsx
+++ b/src/components/dashboard/components/dictionary/common/DictionaryModal.jsx
@@ -23,6 +23,7 @@ export class DictionaryModal extends React.Component {
         public_access: 'View',
         name: '',
         owner: '',
+        custom_validation_schema: 'OpenMRS',
         description: '',
         default_locale: 'en',
         supported_locales: '',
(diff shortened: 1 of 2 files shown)


Karuhanga:
Could you please look into if this behavior is consistent for when someone tries to update a concept with an already existing name?

alexkayabula:
> Could you please look into if this behavior is consistent for when someone tries to **update** a concept with an already existing name?

I have checked it out. There is actually an error. Let me fix it.

alexkayabula:
> > Could you please look into if this behavior is consistent for when someone tries to **update** a concept with an already existing name?
> 
> I have checked it out. There is actually an error. Let me fix it.

The issue is fixed

alexkayabula:
> Could you please look into if this behavior is consistent for when someone tries to **update** a concept with an already existing name?



> Could you please look into if this behavior is consistent for when someone tries to **update** a concept with an already existing name?

This issue is fixed

PR #431, opened 2019-05-07
OCLOMRS-541: Update state persistence to introduce state versioning

Karuhanga:
# JIRA TICKET NAME:
[Update state persistence to introduce state versioning](https://issues.openmrs.org/browse/OCLOMRS-541)

# Summary:
Given that we are persisting state, we have situation where when an individual updates the structure of state, it conflicts with the currently stored state, causing a crash

 

This ticket introduces versioning to solve this issue.

### DIFF
--- a/src/redux/reducers/store.js
+++ b/src/redux/reducers/store.js
@@ -4,21 +4,31 @@ import reduxThunk from 'redux-thunk';
 import logger from 'redux-logger';
 import rootReducer from './index';
 
+export const STORE_VERSION = '1';
+export const CURRENT_STORE_VERSION_KEY = 'currentStoreVersion';
+
 const saveState = (state) => {
   try {
     localStorage.setItem('state', JSON.stringify(state));
+    localStorage.setItem(CURRENT_STORE_VERSION_KEY, STORE_VERSION);
     return undefined;
   } catch (e) {
     return undefined;
   }
 };
 
-const loadState = () => {
+export const loadState = () => {
   try {
     const state = localStorage.getItem('state');
     if (state === null) {
       return undefined;
     }
+
+    const currentStoreVersion = localStorage.getItem(CURRENT_STORE_VERSION_KEY);
+    if (currentStoreVersion !== STORE_VERSION) {
+      return undefined;
+    }
+
     return JSON.parse(state);
   } catch (e) {
     return undefined;
--- a/src/tests/store.test.js
+++ b/src/tests/store.test.js
@@ -1,0 +1,42 @@
+import { loadState, STORE_VERSION, CURRENT_STORE_VERSION_KEY } from '../redux/reducers/store';
+
+describe('loadState', () => {
+  const expectedState = {
+    concepts: [1, 2],
+  };
+
+  beforeEach(() => {
+    localStorage.clear();
+  });
+
+  it('should return undefined when state is null', () => {
+    expect(loadState()).toEqual(undefined);
+  });
+
+  describe('should return undefined when store versions don\'t match', () => {
+    it('should return undefined when store version in local storage is null', () => {
+      localStorage.setItem('state', JSON.stringify(expectedState));
+      expect(loadState()).toEqual(undefined);
+    });
+
+    it('should return undefined when store versions don\'t match', () => {
+      localStorage.setItem(CURRENT_STORE_VERSION_KEY, STORE_VERSION - 1);
+      localStorage.setItem('state', JSON.stringify(expectedState));
+      const loadedState = loadState();
+      expect(loadedState).not.toEqual(expectedState);
+      expect(loadedState).toEqual(undefined);
+    });
+  });
+
+  it('should return the state that was saved when store versions match', () => {
+    localStorage.setItem(CURRENT_STORE_VERSION_KEY, STORE_VERSION);
+    localStorage.setItem('state', JSON.stringify(expectedState));
+    expect(loadState()).toEqual(expectedState);
+  });
+
+  it('should not return the saved state when store versions don\'t match', () => {
+    localStorage.setItem(CURRENT_STORE_VERSION_KEY, STORE_VERSION - 1);
+    localStorage.setItem('state', JSON.stringify(expectedState));
+    expect(loadState()).not.toEqual(expectedState);
+  });
+});


danuluma:
I've noticed that this change introduces another bug. When one logs in then immediately refreshes the page, they're redirected back to the login screen again i.e they are logged out. Kindly look at that.

Karuhanga:
> I've noticed that this change introduces another bug. When one logs in then immediately refreshes the page, they're redirected back to the login screen again i.e they are logged out. Kindly look at that.

Thanks for pointing this out Dan. I happened to be saving the version only the first time it was checked. Updated this to be done when saving the entire state.

src/redux/reducers/store.js
```diff
@@ -4,21 +4,30 @@ import reduxThunk from 'redux-thunk';
...
+export const loadState = () => {
   try {
     const state = localStorage.getItem('state');
     if (state === null) {
       return undefined;
     }
+
+    const storeVersion = localStorage.getItem('storeVersion');
+    if (storeVersion !== STORE_VERSION) {
+      return undefined;
```
danuluma: Now that the storeVersion is being stored simultaneously with the state, its value will always be equal to STORE_VERSION if the state was saved. I don't think this will ever be returned, because there's already an early return in place on line 23 incase there's no state persisted.

Karuhanga: The goal of this PR was to update the structure of the state when it changes in the code. So at the point when the code is loaded, `STORE_VERSION` in the code will be different from the `storeVersion` in local storage. This is the scenario where we want it to be cleared and updated.

danuluma: Okay

src/redux/reducers/store.js
```diff
@@ -4,21 +4,30 @@ import reduxThunk from 'redux-thunk';
 import logger from 'redux-logger';
 import rootReducer from './index';
 
+export const STORE_VERSION = '1';
+
 const saveState = (state) => {
   try {
     localStorage.setItem('state', JSON.stringify(state));
+    localStorage.setItem('storeVersion', STORE_VERSION);
```
dkayiwa: What happens if you change 'storeVersion' to some thing else? In how many places would you have to change it?

Karuhanga: Making this a constant.

src/redux/reducers/store.js
```diff
@@ -4,21 +4,31 @@ import reduxThunk from 'redux-thunk';
 import logger from 'redux-logger';
 import rootReducer from './index';
 
+export const STORE_VERSION = '1';
+export const storeVersion = 'storeVersion';
```
dkayiwa: What is the convention for naming constants?

src/redux/reducers/store.js
```diff
@@ -4,21 +4,31 @@ import reduxThunk from 'redux-thunk';
 import logger from 'redux-logger';
 import rootReducer from './index';
 
+export const STORE_VERSION = '1';
+export const CURRENT_STORE_VERSION = 'currentStoreVersion';
```
dkayiwa: Isn't the current store version equal to 1 instead of 'currentStoreVersion'?

Karuhanga: `STORE_VERSION` is the constant keeping track of the version in the code. `CURRENT_STORE_VERSION` is the constant for retrieving the `current version` in `local storage` and does not change.
Does this make sense?

dkayiwa: What is the name of the parameter that localStorage.getItem takes?

Karuhanga: `currentStoreVersion`

dkayiwa: Is that what the API docs say about the parameter that getItem takes?

Karuhanga: From the docs,  
`The getItem() method of the Storage interface, when passed a key name, will return that key's value, or null if the key does not exist, in the given Storage object.`

Kindly clarify on what I am missing.

dkayiwa: Your constant name should reflect that too.

Karuhanga: Noted. Updated `CURRENT_STORE_VERSION` to `CURRENT_STORE_VERSION_KEY`.
Thanks for pointing that out.

dkayiwa: You have hit the nail on the head. That is exactly what i was looking for. 😊

src/tests/store.test.js
```diff
@@ -0,0 +1,30 @@
...
+      localStorage.setItem(CURRENT_STORE_VERSION_KEY, STORE_VERSION - 1);
+      localStorage.setItem('state', 'state');
+      expect(loadState()).toEqual(undefined);
+    });
+  });
+
+  it('should return the right state that was saved', () => {
+    localStorage.setItem(CURRENT_STORE_VERSION_KEY, STORE_VERSION);
+    localStorage.setItem('state', JSON.stringify('state'));
+    expect(loadState()).toEqual('state');
```
dkayiwa: How do you tell that for instance we have version 1 (that we stored) instead of 2?

Karuhanga: This test uses the same `STORE_VERSION` constant used when retrieving the state, [here](https://github.com/openmrs/openmrs-ocl-client/blob/0bd1f0c1b4154c7eabdf6410a2498d9fb1456249/src/redux/reducers/store.js#L7), so we'd expect the state to be retrieved correctly in this scenario.
The test that checks for a mismatch in versions is on line 18 where we set the version to be one lower than the same constant.

dkayiwa: Let me put it this way, what is the right state which was saved?

Karuhanga: I see.
I have updated the state we check for to be a constant to make this more explicit.
I have also updated the test description.
It would be helpful to point out that `loadState` either returns the state or `undefined` and we do not have a scenario where we have more than one state version in local storage.

Is this okay?

src/tests/store.test.js
```diff
@@ -0,0 +1,33 @@
...
+
+    it('should return undefined when store versions don\'t match', () => {
+      localStorage.setItem(CURRENT_STORE_VERSION_KEY, STORE_VERSION - 1);
+      localStorage.setItem('state', 'state');
+      expect(loadState()).toEqual(undefined);
+    });
+  });
+
+  it('should return the state that was saved when store versions match', () => {
+    localStorage.setItem(CURRENT_STORE_VERSION_KEY, STORE_VERSION);
```
dkayiwa: Can you have another test like this on each and every line where the only difference is? localStorage.setItem(CURRENT_STORE_VERSION_KEY, STORE_VERSION -1)

Karuhanga: The test on line 18([here](https://github.com/openmrs/openmrs-ocl-client/blob/33ff2380ee1679767c34656ea7dd047a687added/src/tests/store.test.js#L22)) tests this same scenario. Changing that line without changing the assertion causes the test to fail, as expected.

dkayiwa: Make the assertion to be the opposite of this one.

Karuhanga: I have added a second assertion to the test on line 22([here](https://github.com/openmrs/openmrs-ocl-client/blob/ab1b18168e553376b2b06a71c1e12d796e5d1acf/src/tests/store.test.js#L22)).
I hope this suffices.

dkayiwa: What i did mean was an opposite of the test that you had before, but with only the assertion changing together with localStorage.setItem(CURRENT_STORE_VERSION_KEY, STORE_VERSION -1)

Karuhanga: Yes, Daniel. What I don't get is, is this not done here already?
```
    it('should return undefined when store versions don\'t match', () => {
      localStorage.setItem(CURRENT_STORE_VERSION_KEY, STORE_VERSION - 1);
      localStorage.setItem('state', JSON.stringify(expectedState));
      const loadedState = loadState();
      expect(loadedState).not.toEqual(expectedState);
      expect(loadedState).toEqual(undefined);
    });
```

dkayiwa: The original test had some data values of 1, 2

Karuhanga: Correct, and we moved these to a single constant, [here](https://github.com/openmrs/openmrs-ocl-client/blob/ab1b18168e553376b2b06a71c1e12d796e5d1acf/src/tests/store.test.js#L4).

dkayiwa: I mean the values you had in expectedState

Karuhanga: The constant is called `expectedState` and still has the same values. I am just reusing it.
image

dkayiwa: Did you just modify the existing test? Because what i did mean was for you to leave the test as it was and create a new one with only those changes in the original test.

Karuhanga: I did not modify the test we were referring to. What I did do was add another asertion to one of the other tests that were testing such a scenario in order to achieve what we were after, which according to this (https://github.com/openmrs/openmrs-ocl-client/pull/431#discussion_r282418513) was to test the opposite scenario. Having yet another test would be redundant.

> Make the assertion to be the opposite of this one.

dkayiwa: First add the test as i had advised and i will tell you what i did mean.

Karuhanga: Done!

dkayiwa: Don't you find this clearer? 😊

Karuhanga: Maybe, but I guess that's a point of perspective 😊.

dkayiwa: May be you can try to convince me. Why would you find the other better?

Karuhanga: Well, IMO I think it's cleaner to have a single test per scenario. Adding the second assertion to the first test makes sense to make it clearer as to what you are testing. The second test doesn't really give us anything the first one isn't!

PR #435, opened 2019-05-10
OCLOMRS-542: Create and edit a concept without a description (which should be optional)

danuluma:
# JIRA TICKET NAME:
[Create and edit a concept without a description (which should be optional)](https://issues.openmrs.org/browse/OCLOMRS-542)

# Summary:
Creating a concept without a description (which should be optional) fails silently. The concept is not created but doesn't notify the user.

Relatedly, creating a concept with a description then editing it to remove the description works. However, on a subsequent edit, trying to remove the empty description added by default causes the app to crash.

### DIFF
--- a/src/components/dictionaryConcepts/components/DescriptionTable.jsx
+++ b/src/components/dictionaryConcepts/components/DescriptionTable.jsx
@@ -1,37 +1,29 @@
-import React from 'react';
+import React, { PureComponent } from 'react';
 import PropTypes from 'prop-types';
 import DescriptionRow from './DescriptionRow';
 
-const DescriptionTable = props => (
-  <table className="table table-striped table-bordered concept-form-table">
-    <thead className="header text-white">
-      <tr>
-        <th scope="col">Description</th>
-        <th scope="col">Language</th>
-        <th scope="col">Actions</th>
-      </tr>
-    </thead>
-    <tbody>
-      {props.existingConcept && props.existingConcept.descriptions
-        ? props.existingConcept.descriptions.map(newRow => (
-          <DescriptionRow
-            newRow={newRow}
-            rowId={newRow.uuid}
-            key={newRow.uuid}
-            {...props}
-          />
-        ))
-        : props.description.map(newRow => (
-          <DescriptionRow
-            newRowUuid={newRow}
-            rowId={newRow}
-            key={newRow}
-            {...props}
-          />))
-      }
-    </tbody>
-  </table>
-);
+class DescriptionTable extends PureComponent {
+  render() {
+    const { existingConcept, description } = this.props;
+    const descriptions = existingConcept.descriptions || description;
+    return (
+      <table className="table table-striped table-bordered concept-form-table">
+        <thead className="header text-white">
+          <tr>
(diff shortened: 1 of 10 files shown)


danuluma:
I'll be creating a separate PR to address the coverage drop since all lines I worked on in this have been fully covered.

src/components/dictionaryConcepts/components/DescriptionTable.jsx
```diff
@@ -1,37 +1,29 @@
...
-            newRowUuid={newRow}
-            rowId={newRow}
-            key={newRow}
-            {...props}
-          />))
-      }
-    </tbody>
-  </table>
-);
+class DescriptionTable extends PureComponent {
```
Karuhanga: Any reason this had to be changed from a functional to a pure component?

src/redux/reducers/util.js
```diff
@@ -70,7 +70,10 @@ export const filterUserPayload = (user, payload) => {
   return filteredDictionaries;
 };
 
-export const filterDescriptions = (item, list) => list.filter(listItem => listItem.uuid !== item);
+export const filterItems = (item, list) => (
```
Karuhanga: Is this the best name for this method given what it does?
Filtering in JS is expected to return items that match something, not the other way around.

src/tests/dictionaryConcepts/components/DescriptionTable.test.jsx
```diff
@@ -30,4 +30,20 @@ describe('Test suite for DescriptionTable', () => {
...
+        uuid: 'newUUID',
+      }, {
+        uuid: 'a.n.other',
+      }],
+    };
+    const wrapper = mount(<DescriptionTable {...newProps} />);
+    expect(wrapper.find('DescriptionRow').length).toEqual(newProps.description.length);
+    expect(wrapper.find('DescriptionRow').at(0).instance().props.newRow)
+      .toEqual(newProps.description[0]);
+    expect(wrapper).toMatchSnapshot();
```
Karuhanga: Given we are updating snapshots each time while testing, is snapshot testing still necessary?

src/components/dictionaryConcepts/components/DescriptionTable.jsx
danuluma: It avoids unnecessary re-rendering of the component - a feature that pure components have which functional components don't. There was a bug where occasionally, edit page would render with more than one empty description row on refreshing the page, and I noticed they had exactly the same props.
Pure components only trigger a re-render if the props passed to it have actually changed

src/redux/reducers/util.js
danuluma: Okay

src/tests/dictionaryConcepts/components/DescriptionTable.test.jsx
danuluma: No, we are not updating snapshots, locally. Check on the tests for other components.

src/redux/reducers/util.js
danuluma: Thanks for that. Changed

PR #437, opened 2019-05-11
OCLOMRS-553: When you edit a concept, a new version of that concept should be created and its reference refreshed in the collection

Karuhanga:
JIRA TICKET NAME:
[When you edit a concept, a new version of that concept should be created and its reference refreshed in the collection](https://issues.openmrs.org/browse/OCLOMRS-553)

# Summary:
- I added 4 concepts from CIEL, I created 4, and I edited one of those 4. In our application, I see that I have 8 concepts (which is what I expect). But in traditional OCL I see 5 references to custom concepts, not 4.
- What I think is happening is that when I edit a concept, you create a new concept version of the old concept, and you also create a new concept, and they’re both included in the collection.
- That’s wrong when I edit it should create a new version of one concept, and refresh its reference in the collection.

### DIFF
--- a/src/redux/actions/dictionaries/dictionaryActionCreators.js
+++ b/src/redux/actions/dictionaries/dictionaryActionCreators.js
@@ -276,3 +276,31 @@ export const retireConcept = (conceptUrl, retired) => async (dispatch) => {
     return null;
   }
 }
+
+export const addReferenceToCollectionAction = (type, owner, collection, expressions) => async (dispatch) => {
+  try {
+    return await api.dictionaries.addReferencesToCollection(type, owner, collection, expressions);
+  } catch (e) {
+    if(e && e.response && e.response.data){
+      notify.show(e.response.data, 'error', 3000);
+    }
+    else{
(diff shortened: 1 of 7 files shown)


alexkayabula:
Nice work @Karuhanga

src/redux/actions/dictionaries/dictionaryActionCreators.js
```diff
@@ -276,3 +276,19 @@ export const retireConcept = (conceptUrl, retired) => async (dispatch) => {
     return null;
   }
 }
+
+export const addReferenceToCollectionAction = (type, owner, collection, expressions) => async (dispatch) => {
+  try {
+    return await api.dictionaries.addReferencesToCollection(type, owner, collection, expressions);
+  } catch (e) {
+    return false;
```
dkayiwa: Isn't the caller interested in knowing the cause of the failure?

Karuhanga: Oversight. Thanks for pointing this out.

PR #438, opened 2019-05-11
OCLOMRS-561: Deleting a reference to an old concept fails sometimes

Karuhanga:
# JIRA TICKET NAME:
[Deleting a reference to an old concept fails sometimes](https://issues.openmrs.org/browse/OCLOMRS-561)

# Summary:
This is an issue with the API and an [issue](https://github.com/OpenConceptLab/ocl_issues/issues/115) has been raised. Pending resolution of this issue, the suggestion is not to display concepts that are not the latest version when viewing a dictionary's concepts.

### DIFF
--- a/src/redux/actions/concepts/dictionaryConcepts.js
+++ b/src/redux/actions/concepts/dictionaryConcepts.js
@@ -137,8 +137,14 @@ export const fetchDictionaryConcepts = (
   }
   try {
     const response = await instance.get(`${url}&includeRetired=1`);
-    dispatch(getDictionaryConcepts(response.data, FETCH_DICTIONARY_CONCEPT));
-    dispatch(paginateConcepts(response.data));
+    // this filter is to account for situations where a ref to a concept
+    // version is not removed after an update. Pending backend fix.
+    // ticket: https://github.com/OpenConceptLab/ocl_issues/issues/115
+    const concepts = response.data.filter(
+      concept => concept.external_id || concept.is_latest_version,
+    );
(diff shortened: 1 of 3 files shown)


src/redux/actions/concepts/dictionaryConcepts.js
```diff
@@ -133,8 +133,13 @@ export const fetchDictionaryConcepts = (
   }
   try {
     const response = await instance.get(`${url}&includeRetired=1`);
-    dispatch(getDictionaryConcepts(response.data, FETCH_DICTIONARY_CONCEPT));
-    dispatch(paginateConcepts(response.data));
+    // this filter is to account for situations where a ref to a concept
```
dkayiwa: Instead of this comment, shouldn't we have a ticket?

Karuhanga: We created a ticket on the main OCL Issues repository([here](https://github.com/OpenConceptLab/ocl_issues/issues/115)). This comment in particular is just to give whoever is working on that part of the code base context as to why this was necessary.
Is it unnecessary in your opinion?

dkayiwa: Fair enough 😊

Karuhanga: Ahh yes. Adding that.

Karuhanga: Added.

PR #421, opened 2019-04-27
OCLOMRS-532: When adding a Q and A CIEL concept, the answer concepts should also be added to the dictionary

myrdstom:
# JIRA TICKET NAME:
[When adding a Q and A CIEL concept, the answer concepts should also be added to the dictionary](https://issues.openmrs.org/browse/OCLOMRS-532)

# Summary:
When adding a Q and A CIEL concept, the answer concepts should also be added to the dictionary

PR #439, opened 2019-05-14
OCLOMRS-533: On the Add CIEL concepts page, when you add a set, it should add its members

myrdstom:
# JIRA TICKET NAME:
[On the Add CIEL concepts page, when you add a set, it should add its members](https://issues.openmrs.org/browse/OCLOMRS-533)

# Summary:
On the Add CIEL concepts page, when you add a set, it should add its members


myrdstom:
It was covered in this [PR](https://github.com/openmrs/openmrs-ocl-client/pull/421/files)

PR #442, opened 2019-05-15
OCLOMRS-559: Changing a datatype when creating and editing a concept is not saved

danuluma:
# JIRA TICKET NAME:
[Changing a datatype when creating and editing a concept is not saved](https://issues.openmrs.org/browse/OCLOMRS-559)

# Summary:
When one creates a concept with a datatype other than 'none', the datatype is not reflected when editing the concept. It always displays 'none'.


danuluma:
Thank you.

PR #445, opened 2019-05-15
OCLOMRS-518: Edit concept should include all mapping types(answers, sets, mappings)

Karuhanga:
# JIRA TICKET NAME:
[Edit concept should have all possible options](https://issues.openmrs.org/browse/OCLOMRS-518)

# Summary:
#### What is expected
The edit concept should display answers, sets and mappings for the concept being edited.

#### Currently
The edit another concept page displays only either set or answer mappings and not both.

### DIFF
--- a/src/components/dictionaryConcepts/components/CreateConceptForm.jsx
+++ b/src/components/dictionaryConcepts/components/CreateConceptForm.jsx
@@ -202,7 +202,7 @@ const CreateConceptForm = (props) => {
             </div>
           </div>
         </div>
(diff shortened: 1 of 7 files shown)


Karuhanga:
> Hi @Karuhanga. I've tested this out and trying to edit a set fails silently. Editing an answer still works well. Kindly look into that.

Thanks for pointing this out @danuluma. As we've discussed though, editing the set without removing it was not implemented yet and a ticket is now up for that.

PR #446, opened 2019-05-15
OCLOMRS-536: Previewing a Concept WIth Both Sets and Answers Does not Show Sets

myrdstom:
# JIRA TICKET NAME:
[Previewing a Concept WIth Both Sets and Answers Does not Show Answers](https://issues.openmrs.org/browse/OCLOMRS-536)

# Summary:
Previewing a concept currently assumes that only an Answer or Sets can be present at a time. This PR caters for the possibility of both Answers and sets being available and displays them both.

### DIFF
--- a/src/components/bulkConcepts/component/PreviewCard.jsx
+++ b/src/components/bulkConcepts/component/PreviewCard.jsx
@@ -1,3 +1,4 @@
+/* eslint-disable no-else-return */
 import React from 'react';
 import PropTypes from 'prop-types';
 import {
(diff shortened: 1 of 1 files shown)


src/components/bulkConcepts/component/PreviewCard.jsx
```diff
@@ -82,6 +82,23 @@ const PreviewCard = ({
         remainingItems={answers.slice(NUMBER_OF_ITEMS_TO_DISPLAY).join(' ')}
       />;
     }
+    if (answers && answers.length > 0 && setMembers && setMembers.length > 0) {
```
dkayiwa: What happens when answers.length > 0 but setMembers.length = 0?

myrdstom: Line 76 handles this scenario. But ideally what would happen is only answer mappings would be displayed

dkayiwa: My question is, if answers.length > 0 but setMembers.length = 0, does this mean you would not render answers?

myrdstom: `if (answers && answers.length > 0 && setMembers && setMembers.length < 1)`
This condition would pass in this scenario, and hence the answers would be displayed. Would it be more satisfactory if I equated the code to something like
(answers && answers.length > 0 && setMembers && setMembers.length = 0?
 My understanding is it is similar to the code I currently have. Or am I missing something else?

dkayiwa: When you say that the condition would pass, which line are you talking about?

myrdstom: Im talking about `line 76` in file `src/components/bulkConcepts/component/PreviewCard.jsx`

dkayiwa: My questions are in reference to line 85

myrdstom: I have refactored the code to use `if else-if` statements. Would that cater for your concern?

dkayiwa: What do your team mates say?

myrdstom: I had afew suggestions, is it possible to give it a final glance?

Karuhanga: > My question is, if answers.length > 0 but setMembers.length = 0, does this mean you would not render answers?



> `if (answers && answers.length > 0 && setMembers && setMembers.length < 1)`
> This condition would pass in this scenario, and hence the answers would be displayed. Would it be more satisfactory if I equated the code to something like
> answers && answers.length > 0 && setMembers && setMembers.length = 0?
> My understanding is it is similar to the code I currently have. Or am I missing something else?

I believe this situation is catered for in the scenario Paul outlines, @dkayiwa. A little clarity on what we might be missing perhaps?

dkayiwa: Why do you combine the check for setmembers and answers in the same if condition?

myrdstom: Thank you for the clarity, I should effect the changes shortly

myrdstom: I have refactored the code please have a look at it when you get some time

PR #450, opened 2019-05-17
OCLOMRS-567: Filter concepts By Class name not working as intended

Karuhanga:
# JIRA TICKET NAME:
[Filter concepts By Class name not working as intended](https://issues.openmrs.org/browse/OCLOMRS-567)

# Summary:
When viewing bulk concepts, the current filter status is not reflected in the checkbox sometimes, causing the functionality to break

### DIFF
--- a/src/tests/bulkConcepts/actions/bulkConcept.test.js
+++ b/src/tests/bulkConcepts/actions/bulkConcept.test.js
@@ -6,7 +6,6 @@ import { notify } from 'react-notify-toast';
 import instance from '../../../config/axiosConfig';
 import {
   IS_FETCHING,
-  FETCH_BULK_CONCEPTS,
   ADD_TO_DATATYPE_LIST,
   ADD_TO_CLASS_LIST,
   FETCH_FILTERED_CONCEPTS,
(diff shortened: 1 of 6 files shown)


Karuhanga:
Raising a PR to account for this coverage drop. Link will be available as soon as it is up.

src/tests/bulkConcepts/actions/bulkConcept.test.js
```diff
@@ -39,48 +37,6 @@ describe('Test suite for addBulkConcepts async actions', () => {
   afterEach(() => {
     moxios.uninstall(instance);
   });
-
-  it('should handle FETCH_BULK_CONCEPTS', () => {
-    moxios.wait(() => {
```
dkayiwa: Could you remind me on how the changes in this file are related to the fix?

Karuhanga: There was a bunch of duplicate code, in particular `fetchBulkConcepts` and `fetchFilteredConcepts` were doing the same thing when filtering was not being done.
Since `fetchBulkConcepts` is only used in this component, it needed to be deleted, along with its test.

PR #451, opened 2019-05-20
WIP-OCLOMRS-571: Raise Test Coverage

Karuhanga:
# JIRA TICKET NAME:
[Raise Test Coverage](https://issues.openmrs.org/browse/OCLOMRS-571)

# Summary:
Address the falling coverage on OCLOMRS-567

### DIFF
--- a/src/tests/redux/reducers/util.test.js
+++ b/src/tests/redux/reducers/util.test.js
@@ -1,0 +1,21 @@
+import { filterClass, filterList, filterSources } from '../../../redux/reducers/util';
+
+const arg = 'arg';
+
+describe('filterSources', () => {
+  it('should return an empty list if its argument is not an array', () => {
+    expect(filterSources(arg)).toEqual([]);
+  });
+});
+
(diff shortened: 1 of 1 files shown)


src/tests/redux/reducers/util.test.js
```diff
@@ -0,0 +1,21 @@
+import { filterClass, filterList, filterSources } from '../../../redux/reducers/util';
+
+const arg = 'arg';
+
+describe('filterSources', () => {
+  it('should return an empty list if its argument is not an array', () => {
```
dkayiwa: Does this mean the argument can be anything?
Do we have a test for when the argument is as expected?

Karuhanga: > Do we have a test for when the argument is as expected?

I had assumed so, since the lines were being covered, but looking through the code base, it appears that may not be the case.
Adding these tests.

Karuhanga:
Closing this temporarily until all high priority issues are addressed

PR #454, opened 2019-05-22
OCLOMRS-573: Set auto focus on the dictionary name field when a creating new dictionary

danuluma:
# JIRA TICKET NAME:
[Set auto focus on the dictionary name field when a creating new dictionary](https://issues.openmrs.org/browse/OCLOMRS-573)

# Summary:
When creating a dictionary the default cursor position should be set in the dictionary name field.

### DIFF
--- a/src/components/dashboard/components/dictionary/common/DictionaryModal.jsx
+++ b/src/components/dashboard/components/dictionary/common/DictionaryModal.jsx
@@ -33,6 +33,7 @@ export class DictionaryModal extends React.Component {
       errors: {},
       disableButton: false,
     };
+    this.nameInput = undefined;
   }
 
   componentDidMount() {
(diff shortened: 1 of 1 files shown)


dkayiwa:
What happened to the autofocus attribute?

danuluma:
> What happened to the autofocus attribute?

Apparently, this was not working as expected.

dkayiwa:
Did you investigate why?

danuluma:
> Did you investigate why?

Yes, the reason it did not work was that we are displaying the form in a modal, which makes the modal to be the parent component. Using the `autoFocus` attribute sets the focus on the modal(parent) instead of the input field(which is in a child component). The way this could be done was by using a `ref` to manually accessing the form and set `focus` on the required field. Refs, however, work only on class-based components and we are using react-strap, whose inputs(or children components) are functional components. However, again, they have created a  custom attribute `innerRef` which can be used to access the child components in react-strap form. This is what I made use of to access the name input and set focus on it.

[Reference](https://github.com/reactstrap/reactstrap/issues/1021)

PR #455, opened 2019-05-22
OCLOMRS-574: Internal Mappings Not Created When Creating or Updating a Concept

Karuhanga:
# JIRA TICKET NAME:
[Internal Mappings Not Created When Creating or Updating a Concept](https://issues.openmrs.org/browse/OCLOMRS-574)

# Summary:
The body of data sent to the API when creating mappings to internal concepts is incorrect.

### DIFF
--- a/src/tests/dictionaryConcepts/actions/dictionaryConcept.test.js
+++ b/src/tests/dictionaryConcepts/actions/dictionaryConcept.test.js
@@ -79,6 +79,7 @@ import {
   prePopulateSets,
   unpopulatePrepopulatedSets,
   unpopulateSet,
+  buildNewMappingData,
 } from '../../../redux/actions/concepts/dictionaryConcepts';
 import {
   removeDictionaryConcept,
@@ -94,7 +95,12 @@ import concepts, {
   newConceptDataWithAnswerAndSetMappings,
   existingConcept, sampleConcept, conceptWithoutMappings,
 } from '../../__mocks__/concepts';
-import { CIEL_SOURCE_URL, INTERNAL_MAPPING_DEFAULT_SOURCE, MAP_TYPE } from '../../../components/dictionaryConcepts/components/helperFunction';
+import {
+  CIEL_SOURCE_URL,
+  INTERNAL_MAPPING_DEFAULT_SOURCE,
+  MAP_TYPE,
+  MAP_TYPES_DEFAULTS
+} from '../../../components/dictionaryConcepts/components/helperFunction';
 
 jest.mock('uuid/v4', () => jest.fn(() => 1));
 jest.mock('react-notify-toast');
@@ -504,6 +510,29 @@ describe('Test suite for dictionary concept actions', () => {
       expect(store.getActions()).toEqual(expectedActions);
     });
   });
+
+  it('should handle any unknown error in createNewConcept', () => {
+    const notifyMock = jest.fn();
+    notify.show = notifyMock;
+
+    moxios.wait(() => {
+      const request = moxios.requests.mostRecent();
+      request.reject({});
+    });
+
+    const expectedActions = [
+      { type: IS_FETCHING, payload: true },
+      { type: IS_FETCHING, payload: false },
+    ];
+
+    const store = mockStore(mockConceptStore);
+    const url = '/orgs/IHTSDO/sources/SNOMED-CT/concepts/';
+    return store.dispatch(createNewConcept(newConceptData, url)).then(() => {
+      expect(store.getActions()).toEqual(expectedActions);
+      expect(notifyMock).toHaveBeenCalledWith('An error occurred when creating a concept. Please retry.', 'error', 2000);
+    });
+  });
+
   it('should handle ADD_CONCEPT_TO_DICTIONARY', () => {
     moxios.wait(() => {
       const request = moxios.requests.mostRecent();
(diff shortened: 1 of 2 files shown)


src/tests/dictionaryConcepts/actions/dictionaryConcept.test.js
```diff
@@ -1187,3 +1192,84 @@ describe('Add set mappings to concept', () => {
...
+      map_type: mapType,
+      from_concept_url: fromConceptUrl,
+      to_source_url: toSourceUrl,
+      to_concept_code: toConceptCode,
+      to_concept_name: toConceptName,
+    };
+    expect(buildNewMappingData(mapping, fromConceptUrl)).toEqual(expectedMapping);
+  });
+
+  it('should return the right mapping data if given an internal concept', () => {
```
dkayiwa: How different is this from the last test?

Karuhanga: The source for the mapping is different and hence different data is returned as evidenced by the differing expectedMappings. Line 1205 and Line 1223.

dkayiwa: Is that why the description is the same?

Karuhanga: The description is not the same.
> should return the right mapping data if given an internal concept
> should return the right mapping data if given an external concept

dkayiwa: should return the right mapping data if given an internal concept
should return the right mapping data if given an internal concept

Karuhanga: Are we looking at the same file? 🙂
image

dkayiwa: https://github.com/openmrs/openmrs-ocl-client/pull/455/files

Karuhanga: Lines 1231 and 1249 are not the same Daniel, are they?
Could you kindly point out the line numbers, perhaps therein lies the confusion.

dkayiwa: 1249, 1289

Karuhanga: Thank you. These tests are each for different methods(and are in different describe blocks). One for `buildUpdateMappingData` and the other for `buildNewMappingData` which return distinct types of data.

dkayiwa: Why do they have the same description is they are testing different things?

Karuhanga: The tests on 1249 and 1289 test that those methods `should return the right mapping data if given an internal concept`. The meaning of `right mapping data` defers for each of those methods but that name still stands. Perhaps if you took a look at what those methods do this would be more clear?

They are testing the ~same thing on different methods.

dkayiwa: The fact that am asking this should trigger in your mind that latter maintainers of this code will struggle to understand why you used the same description for different things.

Karuhanga: Noted. Thanks for pointing this out. Updated these descriptions to explicitly state what each is testing.

src/tests/dictionaryConcepts/actions/dictionaryConcept.test.js
```diff
@@ -1214,3 +1269,84 @@ describe('Add set mappings to concept', () => {
...
+    const expectedMapping = {
+      map_type: mapType,
+      to_source_url: toSourceUrl,
+      to_concept_code: toConceptCode,
+      to_concept_name: toConceptName,
+    };
+    expect(buildNewMappingData(mapping)).toEqual(expectedMapping);
+  });
+
+  it('should return the map_type, to_concept_url, to_concept_name if given an internal concept', () => {
```
dkayiwa: I have seen another "if given an internal concept" above. Why do these tests return different values?

Karuhanga: Because they are testing different methods.

dkayiwa: How do i tell that?

Karuhanga: They are in different describe blocks Daniel. Is this not descriptive enough?

dkayiwa: Have you thought of maintainers of this code would will be asking the same questions like me?

Karuhanga: I have updated these descriptions to include the names of the methods they are testing.

PS: I am not convinced that this makes it hard to understand the goal of the tests because, for example when jest displays failing tests, it displays the full context including under which describe blocks a test lies.
This in my opinion goes against the convention for structuring tests. As always, I'd love to be convinced otherwise. 🙂

dkayiwa: Would you rather that i first look at the body of a test to know what is exactly being tested? Or know it all by just looking at the description? 😊

Karuhanga: The latter, true.
But, I believe in this situation the location should be taken into context as well, that is which file and which describe block, don't you think?

dkayiwa: You have a good point in regards to that. So am convinced and merging. 😊

PR #468, opened 2019-05-30
OCLOMRS-583: Dictionary Summary Page shows 'Invalid Date' after  releasing a new collection version

danuluma:
# JIRA TICKET NAME:
[Dictionary Summary Page shows 'Invalid Date' after releasing a new collection version](https://issues.openmrs.org/browse/OCLOMRS-583)

# Summary:
Dictionary Summary page shows an "Invalid date" immediately after releasing a new collection version.

### DIFF
--- a/src/components/dashboard/components/dictionary/DictionaryContainer.jsx
+++ b/src/components/dashboard/components/dictionary/DictionaryContainer.jsx
@@ -63,7 +63,7 @@ export class DictionaryOverview extends Component {
     this.props.fetchDictionaryConcepts(conceptsUrl);
   }
 
-  componentWillUpdate(prevProps) {
+  componentDidUpdate = (prevProps) => {
     const {
       match: {
         params: {
@@ -73,7 +73,7 @@ export class DictionaryOverview extends Component {
     } = this.props;
     const versionUrl = `/${ownerType}/${owner}/${type}/${name}/versions/?verbose=true`;
 
-    if (prevProps.isReleased !== this.props.isReleased) {
+    if (prevProps.versions.length !== this.props.versions.length) {
       this.props.fetchVersions(versionUrl);
     }
   }
--- a/src/tests/Dictionary/DictionaryContainer.test.jsx
+++ b/src/tests/Dictionary/DictionaryContainer.test.jsx
@@ -155,7 +155,7 @@ describe('DictionaryOverview', () => {
     expect(spy).toHaveBeenCalledTimes(1);
   });
 
-  it('should handle componentWillUpdate in HEAD version release', () => {
+  it('should fetch all released versions when a new collection version is released', () => {
     const props = {
       dictionary,
       versions: [customVersion],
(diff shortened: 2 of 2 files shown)


src/components/dashboard/components/dictionary/DictionaryContainer.jsx
```diff
@@ -73,7 +73,7 @@ export class DictionaryOverview extends Component {
     } = this.props;
     const versionUrl = `/${ownerType}/${owner}/${type}/${name}/versions/?verbose=true`;
 
-    if (prevProps.isReleased !== this.props.isReleased) {
```
dkayiwa: What has this got to do with the invalid date?

src/tests/Dictionary/DictionaryContainer.test.jsx
```diff
@@ -155,7 +155,7 @@ describe('DictionaryOverview', () => {
     expect(spy).toHaveBeenCalledTimes(1);
   });
 
-  it('should handle componentWillUpdate in HEAD version release', () => {
+  it('should fetch all released versions when a new collection version is released', () => {
```
dkayiwa: What has this got to do with the invalid date?

src/components/dashboard/components/dictionary/DictionaryContainer.jsx
danuluma: Here, the action `fetchVersions` (which fetches all the details of released versions - including the release date) needs to be triggered whenever a new version is released. However, the property `isReleased` changes only once - when the first version is released. Since we need to call `fetchVersions` on every new release, monitoring a change in this won't work because its value will only change on the first release. 

The number of versions released, however, increases with every release. I, therefore, changed the logic to be monitoring this and call `fetchVersions` whenever there is a change - which would occur every time a new collection version is released.

src/tests/Dictionary/DictionaryContainer.test.jsx
danuluma: I had to re-write this test and test the changed functionality.

dkayiwa: I still do not understand how it is related to a ticket which is dealing with the display of an invalid date. The only test that i would expect in such a ticket would be "it should display a valid date...."

src/components/dashboard/components/dictionary/DictionaryContainer.jsx
dkayiwa: Is that what causes the display of an invalid date?

danuluma: Yes,  because the value of the date would be null - which can't be formated hence 'invalid date' would be returned.

src/tests/Dictionary/DictionaryContainer.test.jsx
danuluma: The root cause of the 'invalid date' bug was that the released collections' data was not being updated (re-fetched from the backend) after a new release. That's what I actually changed and is what this is testing for. Displaying 'invalid date' was a result of the data not being fetched.

PR #470, opened 2019-05-31
OCLOMRS-591: When adding a concept to a collection, its mappings should be added as well

Karuhanga:
# JIRA TICKET NAME:
[When adding a concept to a collection, it's mappings should be added as well](https://issues.openmrs.org/browse/OCLOMRS-591)

# Summary:
The endpoint for adding concept references to a collection includes a cascade query option.

We want to add these mappings to the collection, so it should be set to cascade=sourcemappings

### DIFF
--- a/src/components/bulkConcepts/addBulkConcepts.jsx
+++ b/src/components/bulkConcepts/addBulkConcepts.jsx
@@ -72,7 +72,7 @@ export class AddBulkConcepts extends Component {
         params: { type, typeName, collectionName },
       },
     } = this.props;
-    const url = `${type}/${typeName}/collections/${collectionName}/references/`;
+    const url = `${type}/${typeName}/collections/${collectionName}/references/?cascade=sourcemappings`;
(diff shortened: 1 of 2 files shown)


src/components/bulkConcepts/addBulkConcepts.jsx
```diff
@@ -72,7 +72,7 @@ export class AddBulkConcepts extends Component {
         params: { type, typeName, collectionName },
       },
     } = this.props;
-    const url = `${type}/${typeName}/collections/${collectionName}/references/`;
+    const url = `${type}/${typeName}/collections/${collectionName}/references/?cascade=sourcemappings`;
```
dkayiwa: What is the use of / before ?

Karuhanga: The API endpoints expect trailing slashes. It returns a `500` when there's no slash.

PR #472, opened 2019-06-02
OCLOMRS-589: There should be a separate field for performing a search for internal mappings when creating or updating them

Karuhanga:
# JIRA TICKET NAME:
[There should be a separate field for performing a search for internal mappings when creating or updating them](https://issues.openmrs.org/browse/OCLOMRS-589)

# Summary:
Currently, the same fields used to enter the mapping details are used for search. This is not right, and the search box should be an independent field.


danuluma:
Please address the drop in test coverage

PR #480, opened 2019-06-07
OCLOMRS-599: A loader should be displayed when adding CIEL concepts recursively until they are all fully added

danuluma:
# JIRA TICKET NAME:
[OCLOMRS-599](https://issues.openmrs.org/browse/OCLOMRS-599)

# Summary:
There should be an indicator when adding CIEL concepts recursively until they are all fully added


danuluma:
> The changes you made in the action should also be made here `src/redux/actions/bulkConcepts/index.js`
> 
> Also, Is the Jiira ticket name sufficient?

Working on that. Thanks

danuluma:
The drop in coverage has been addressed on #485

PR #481, opened 2019-06-07
OCLOMRS-600: Creating mappings to internal concepts should all have a single searchable field

Karuhanga:
# JIRA TICKET NAME:
[Creating mappings to internal concepts should all have a single searchable field](https://issues.openmrs.org/browse/OCLOMRS-600)

# Summary:
- Creating mappings to internal concepts should all have only one single search field
- Currently, there are three fields, which can be confusing to the user


Karuhanga:
Coverage drop addressed by https://github.com/openmrs/openmrs-ocl-client/pull/482.

PR #486, opened 2019-06-10
OCLOMRS-607: The Add a Ciel Concept page occasionally returns a single value

myrdstom:
# JIRA TICKET NAME:
[The Add a Ciel Concept page occasionally returns a single value](https://issues.openmrs.org/browse/OCLOMRS-607)

# Summary:
The Add a Ciel Concept page occasionally returns a single value

### DIFF
--- a/src/components/bulkConcepts/component/ConceptTable.jsx
+++ b/src/components/bulkConcepts/component/ConceptTable.jsx
@@ -24,7 +24,8 @@ const ConceptTable = ({
         currentPage={currentPage}
         loading={loading}
         PaginationComponent={ConceptPagination}
-        defaultPageSize={concepts.length <= conceptLimit ? concepts.length : conceptLimit}
+        defaultPageSize={conceptLimit}
+        pageSize={concepts.length <= conceptLimit ? concepts.length : conceptLimit}
         noDataText="No concepts found!"
         minRows={2}
         limitCount={10}


src/components/bulkConcepts/component/ConceptTable.jsx
```diff
@@ -24,7 +24,7 @@ const ConceptTable = ({
         currentPage={currentPage}
         loading={loading}
         PaginationComponent={ConceptPagination}
-        defaultPageSize={concepts.length <= conceptLimit ? concepts.length : conceptLimit}
```
dkayiwa: Do you happen to know why this code was originally written as so?

myrdstom: Seen... and shall revert

myrdstom: I have made some changes to the code. Please share your thoughts

myrdstom: I have made some changes to the code. Please share your thoughts

dkayiwa: How was it able to work with your original changes?

myrdstom: Looking through both the code on the `ConceptTable` page and `BulkConceptsPage`, I realized that the defaultPageSize being occasionally set to  1 and unmounted() made it render with the same value when the component mounts again.

The quick fix was to adjust the defaultPageSize to a fixed value of 10 as the functionality for filtering of concepts when searching is handled by the `searchOption` function in the `BulkConceptsPage` making the ternary on `defaultPageSize` redundant.

dkayiwa: Why a value of 10?

myrdstom: The POs have always been of the view that when rendering the `Add CIEL Concepts` page, the user would see 10 CIEL concepts at a time so the `conceptLimit`. This is further addressed in `OCLOMRS-316` variable was created in November 2018 and equated to 10.

dkayiwa: But i did not see 10 in your original changes

myrdstom: You saw `conceptLimit` in my original changes which is set to 10 in `bulkConceptsPage: line 57` and then imported into `ConceptTable` on `line 10`

PR #497, opened 2019-06-14
OCLOMRS-615: Filtering and searching at the same time is not working on the dictionary concepts page

Karuhanga:
# JIRA TICKET NAME:
[Filtering and searching at the same time is not working on the dictionary concepts page](https://issues.openmrs.org/browse/OCLOMRS-615)

# Summary:
Using one of these features turns off the other. Ideally a user should be able to search through filtered concepts

### DIFF
--- a/src/tests/dictionaryConcepts/actions/dictionaryConcept.test.js
+++ b/src/tests/dictionaryConcepts/actions/dictionaryConcept.test.js
@@ -44,7 +44,6 @@ import {
 } from '../../../redux/actions/types';
 import {
   fetchDictionaryConcepts,
-  fetchConceptsByName,
   filterByClass,
   filterBySource,
   createNewName,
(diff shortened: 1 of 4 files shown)


src/tests/dictionaryConcepts/actions/dictionaryConcept.test.js
```diff
@@ -317,48 +316,6 @@ describe('Test suite for dictionary concept actions', () => {
     });
   });
 
-  it('should fetch dictionary concepts by name', () => {
-    moxios.wait(() => {
-      const request = moxios.requests.mostRecent();
```
dkayiwa: Are these tests removed because they were not correct?

Karuhanga: No, they are removed because I deleted some duplicate code. Particularly [here](https://github.com/openmrs/openmrs-ocl-client/pull/497/files#diff-d932a1b328dd7277dbdcffb6b4d1a561).

PR #498, opened 2019-06-15
OCLOMRS-606: Logging out on one tab should log the user out in all other tabs and windows

danuluma:
# JIRA TICKET NAME:
[Logging out on one tab should log the user out in all other tabs and windows](https://issues.openmrs.org/browse/OCLOMRS-606)

# Summary:
When one logs out on one tab, they are not logged out from other tabs and can still try to perform actions that require authentication. This fails and sometimes causes the app to crash.

This ticket is to fix this and make it such that when one logs outs, they are automatically logged out from all the other windows and tabs automatically. Also, when one logs in, they'll be automatically logged in to any open windows and/or tabs and be able to make authenticated requests.

### DIFF
--- a/src/components/Auth/index.jsx
+++ b/src/components/Auth/index.jsx
@@ -1,12 +1,28 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
+import { notify } from 'react-notify-toast';
+import history from '../../config/history';
 
-export const mapStateToProps = ({ users: { loggedIn } }) => ({
-  loggedIn,
-});
+export const checkAuth = (event) => {
+  if (event.key === null) {
+    const token = localStorage.getItem('token');
+    if (!token) {
+      notify.show('You have been logged out', 'warning', 3000);
+      history.go(0);
+    }
+  }
+  if (event.key === 'token') {
+    const token = localStorage.getItem('token');
+    if (token) {
+      notify.show('Logged in successfully', 'success', 3000);
+      history.go(0);
+    }
+  }
+};
 
 export const AuthWrapper = (WrappedComponent) => {
+  window.addEventListener('storage', event => checkAuth(event));
   class Authenticate extends Component {
     static propTypes = {
       loggedIn: PropTypes.bool,
@@ -41,4 +57,8 @@ export const AuthWrapper = (WrappedComponent) => {
   return Authenticate;
 };
 
-export default Authenticate => connect(mapStateToProps, null)(AuthWrapper(Authenticate));
+export const mapStateToProps = ({ users: { loggedIn } }) => ({
+  loggedIn,
+});
+
+export default Authenticate => connect(mapStateToProps)(AuthWrapper(Authenticate));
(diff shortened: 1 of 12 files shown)


src/tests/Dashboard/action/dictionaryAction.test.js
```diff
@@ -185,6 +187,21 @@ describe('Test suite for dictionary actions', () => {
     });
   });
 
+  it('should alert a user and cancel the request if not logged in', () => {
+    const notifyMock = jest.fn();
+    notify.show = notifyMock;
+    localStorage.clear();
+    moxios.wait(() => {
+      moxios.requests.mostRecent();
```
Karuhanga: Are we doing anything with this request?

src/components/Auth/index.jsx
```diff
@@ -1,12 +1,29 @@
...
 
-export const mapStateToProps = ({ users: { loggedIn } }) => ({
-  loggedIn,
-});
+export const checkAuth = (event) => {
+  if (event.key === null) {
+    const token = localStorage.getItem('token');
+    if (!token) {
+      logoutAction();
+      history.go(0);
```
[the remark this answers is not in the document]
danuluma: `history.go(0)` simply reloads the current page, which is the behaviour intended. Its similar to plain js `window.location.reload()`

The `loggedIn` status is stored in the redux store,,which is **not** shared across tabs and windows. Logging out in one window does not change this value in other windows and tabs hence cannot be relied upon

src/tests/Dashboard/action/dictionaryAction.test.js
danuluma: Nothing, it's redundant. Fixed,, thank you

src/components/Auth/index.jsx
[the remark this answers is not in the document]
Karuhanga: I think `history.go(0)` goes to the first index in the history, which might not be the current page.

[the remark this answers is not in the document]
Karuhanga: Anyway, what I'm wondering is, isn't dispatching the logged out action enough?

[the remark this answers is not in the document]
danuluma: `history.go(0)` goes to the first index in the history array. This is **always** the current page.....FYI positive numbers will go forward and negative numbers backwards, relative to the current location (0)

danuluma:
The drop in test coverage has been addressed in #499

src/redux/reducers/authReducers.js
```diff
@@ -30,14 +29,13 @@ export default (state = initialState, action) => {
         loading: false,
       };
     case AUTHENTICATION_FAILED:
```
Karuhanga: When authentication fails, does this mean we don't change state?

danuluma: Yes we are still changing the state same way as before,,

danuluma: Same as logging out,,just tied these two together

Karuhanga: Ahh. Missed that. Cool

PR #519, opened 2019-07-03
OCLOMRS-663: Improve color coherence across application and move all app colors to single file

Karuhanga:
# JIRA TICKET NAME:
[Improve color coherence across application and move all app colors to single file](https://issues.openmrs.org/browse/OCLOMRS-663)

# Summary:
The application currently has multiple incoherent colors from page to page
The goal here is to cut this down to three colors max for a more coherent experience
This ticket also creates color variables , allowing for a single source of truth for all colors in the application

### DIFF
--- a/src/components/dictionaryConcepts/components/ActionButtons.jsx
+++ b/src/components/dictionaryConcepts/components/ActionButtons.jsx
@@ -23,6 +23,14 @@ const ActionButtons = ({
     <React.Fragment>
       {source !== CUSTOM_SOURCE && (
         <React.Fragment>
+          <a
+            href={`${TRADITIONAL_OCL_HOST}${url}`}
+            target="_blank"
+            rel="noopener noreferrer"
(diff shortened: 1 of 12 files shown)


src/components/dictionaryConcepts/components/ActionButtons.jsx
```diff
@@ -23,6 +23,14 @@ const ActionButtons = ({
     <React.Fragment>
       {source !== CUSTOM_SOURCE && (
         <React.Fragment>
+          <a
+            href={`${TRADITIONAL_OCL_HOST}${url}`}
+            target="_blank"
```
dkayiwa: Is this a formatting change?

Karuhanga: Technically, no, but it is a field swap so that we have the remove and retire buttons are in the same column.
New ticket?

PR #522, opened 2019-07-09
OCLOMRS-665: Remove verbose call while fetching collections on the home page

Karuhanga:
# JIRA TICKET NAME:
[Remove verbose call while fetching collections on the home page](https://issues.openmrs.org/browse/OCLOMRS-665)

# Summary:
We currently retrieve the list of collections with a collection's entire details.

As users use the application more, it is possible that they could have a number of collections to manage (>20)

We currently do not(for the user's collections- can be implemented and ticket is up)and cannot(for public dictionaries- because we do some frontend filtering necessitating a back end improvement) have pagination implemented for when we fetch these collections.

This ticket therefore eliminates this verbose calls to prevent this performance degradation.

### DIFF
--- a/src/components/dashboard/components/dictionary/DictionaryCard.jsx
+++ b/src/components/dashboard/components/dictionary/DictionaryCard.jsx
@@ -7,8 +7,6 @@ const DictionaryCard = (dictionary) => {
       name,
       owner,
       owner_type,
-      active_concepts,
-      created_by,
       url,
     },
     gotoDictionary,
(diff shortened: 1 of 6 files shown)


Karuhanga:
This ticket has been put on hold. [This comment](https://issues.openmrs.org/browse/OCLOMRS-665?focusedCommentId=257848&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-257848) on the ticket on JIRA goes onto the rationale.

PR #541, opened 2019-08-01
OCLOMRS-695: Change  to Symptom/Finding in all cases

myrdstom:
# JIRA TICKET NAME:
[Change Symptom-Finding to Symptom/Finding in all areas](https://issues.openmrs.org/browse/OCLOMRS-695)

# Summary:
The 'Symptom/Finding' class is misappropriately named to 'Symptom-Finding'. This is preventing the creation of custom 'Symptom/Finding' concepts. This PR fixes this issue

### DIFF
--- a/src/components/dictionaryConcepts/components/ActionButtons.jsx
+++ b/src/components/dictionaryConcepts/components/ActionButtons.jsx
@@ -19,6 +19,7 @@ const ActionButtons = ({
   if (actionButtons === true) {
     showExtra = true;
   }
+  const conceptClass = (concept_class === 'Symptom/Finding' ? 'Symptom-Finding' : concept_class);
   return (
     <React.Fragment>
       {source !== CUSTOM_SOURCE && (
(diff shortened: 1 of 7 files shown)


dkayiwa:
@Karuhanga did you take a look at this?

Karuhanga:
> @Karuhanga did you take a look at this?

I did and provided one on one feedback, which has been addressed. I'll be approving after I take a look at the before and after videos(Login to JIRA is currently down).

dkayiwa:
Where is Symptom-Finding coming from?

myrdstom:
> Where is Symptom-Finding coming from?

It was initially created as a class option instead of `Symptom/Finding`. This was an error on our part as the Andela team. As a result, we have not been able to create `Symptom/Finding` custom concepts. 

We noticed this during the testing phase. and the origin of this ticket is part of the discussion from [this](https://talk.openmrs.org/t/developing-the-ocl-for-openmrs-application/17771/1162?u=paulk ) talk thread.

dkayiwa:
Can we just change the initial value to? Symptom/Finding

Karuhanga:
> Can we just change the initial value to? Symptom/Finding

That would be ideal, yes but since we currently use the concept class in the url, there's potential for conflict if we use something with a slash.

myrdstom:
> Can we just change the initial value to? Symptom/Finding

That would be ideal, yes but since we currently use the concept class in the url, there's potential for conflict if we use something with a slash.

PR #545, opened 2019-08-06
OCLOMRS-698: A user should be able to bring a newer concept version into their collection to replace the older version

Karuhanga:
# JIRA TICKET NAME:
[A user should be able to bring a newer concept version into their collection to replace the older version](https://issues.openmrs.org/browse/OCLOMRS-698)

# Summary:
Currently, the application does not provide a way for users to update concepts in their collections that may not be the latest version.


Karuhanga:
Closing this for the moment as the discussion into how best to approach this is still ongoing.

PR #548, opened 2019-08-14
OCLOMRS-531: Search results on the Add CIEL concepts page should be consistent, with the best match by name and ID appearing first

Karuhanga:
# JIRA TICKET NAME:
[Search results on the Add CIEL concepts page should be consistent, with the best match by name and ID appearing first](https://issues.openmrs.org/browse/OCLOMRS-531)

# Summary:
- I searched for 1268 (which I actually already added to my dictionary, and the first results are 126800
- This is bad behaviour, but I don’t know which it is:
- If something is already in the dictionary, we should still show it in the search results. (Okay to grey it out, or disable the add button.)
- Exact match on concept ID should show be the first result. This may be a back-end fix, but if this is the scenario, please create a ticket to track it. Make it an MVP ticket for now, though we could punt it.

### DIFF
--- a/src/redux/actions/concepts/addBulkConcepts/index.js
+++ b/src/redux/actions/concepts/addBulkConcepts/index.js
@@ -33,7 +33,7 @@ export const fetchFilteredConcepts = (source = 'CIEL', query = '', currentPage =
   } = getState();
 
   const searchQuery = buildPartialSearchQuery(query);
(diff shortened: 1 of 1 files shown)


Karuhanga:
Work on this ticket would require that the `bestMatch` sort criteria is functioning correctly. https://github.com/openmrs/openmrs-ocl-client/pull/552 works on the result consistency for now.

Karuhanga:
Work on this ticket would require that the `bestMatch` sort criteria is functioning correctly. https://github.com/openmrs/openmrs-ocl-client/pull/552 works on the result consistency for now.

PR #552, opened 2019-08-15
OCLOMRS-705: Search results on the Add CIEL concepts page should be consistent regardless of any action take by the user

Karuhanga:
# JIRA TICKET NAME:
[Search results on the Add CIEL concepts page should be consistent regardless of any action take by the user](https://issues.openmrs.org/browse/OCLOMRS-705)

# Summary:
Addressing part of [OCLOMRS-705](https://issues.openmrs.org/browse/OCLOMRS-705)
- If something is already in the dictionary, we should still show it in the search results. (Okay to grey it out, or disable the add button.)


Karuhanga:
This will be handled by https://issues.openmrs.org/browse/OCLOMRS-719.

PR #553, opened 2019-08-21
OCLOMRS-703: Remove isErrored action creator

judeinno:
# JIRA TICKET NAME:
[Remove isErrored action creator](https://issues.openmrs.org/browse/OCLOMRS-703)

# Summary:
Some parts of the code use isErrored to dispatch an error response into state, causing unexpected crashes. These parts should be refactored to notify the user of the error, but not add it to state.

Dispatching a zero value(eg. empty list) would make sense.

### DIFF
--- a/src/redux/actions/bulkConcepts/index.js
+++ b/src/redux/actions/bulkConcepts/index.js
@@ -2,7 +2,6 @@ import { notify } from 'react-notify-toast';
 import instance from '../../../config/axiosConfig';
 import {
   isSuccess,
-  isErrored,
   isFetching,
   clear,
 } from '../globalActionCreators';
@@ -10,12 +9,12 @@ import {
   ADD_EXISTING_BULK_CONCEPTS,
   FETCH_CONCEPT_SOURCES,
   CLEAR_SOURCE_CONCEPTS,
-  IS_LOADING,
 } from '../types';
 import { recursivelyFetchConceptMappings } from '../concepts/addBulkConcepts';
 import { MAPPINGS_RECURSION_DEPTH } from '../../../components/dictionaryConcepts/components/helperFunction';
 import { deleteNotification, upsertNotification } from '../notifications';
 import { ADDING_CONCEPTS_WARNING_MESSAGE } from '../../../constants';
+import { checkErrorMessage } from '../../../helperFunctions';
 
 export const addExistingBulkConcepts = conceptData => async (dispatch) => {
   const { url, data, conceptIdList: fromConceptIds } = conceptData;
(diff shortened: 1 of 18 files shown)


src/redux/actions/bulkConcepts/index.js
```diff
@@ -26,7 +25,9 @@ const fetchSourceConcepts = url => async (dispatch) => {
     dispatch(isSuccess(response.data, FETCH_SOURCE_CONCEPTS));
     dispatch(isSuccess(false, IS_LOADING));
   } catch (error) {
-    dispatch(isErrored(error.response.data, FETCH_SOURCE_CONCEPTS));
+    if (error && error.response) {
+      notify.show('Failed to fetch source concepts', 'error', 3000);
```
Karuhanga: Any reason we don't make use of the actual error message? I think the convention was to check if there is an error message we can show and show it, or show this generic error otherwise.

src/redux/actions/concepts/dictionaryConcepts.js
```diff
@@ -436,9 +436,8 @@ export const fetchExistingConcept = conceptUrl => async (dispatch) => {
     dispatch(prepopulateAnswers(answers));
     dispatch(prePopulateSets(sets));
   } catch (error) {
-    notify.show('Could not retrieve concept details', 'error', 3000);
-    if (error.response) {
-      dispatch(isErrored(error.response.data, FETCH_EXISTING_CONCEPT_ERROR));
```
Karuhanga: Is there any reason we can't also get rid of `FETCH_EXISTING_CONCEPT_ERROR` since it is not really helping with anything?

In the reducer;
```
case FETCH_EXISTING_CONCEPT_ERROR:
      return {
        ...state,
      };
```

src/redux/actions/bulkConcepts/index.js
judeinno: I have modified it to check for error messages before it displays a custom one.

src/redux/actions/concepts/dictionaryConcepts.js
judeinno: I have removed it since its nolonger needed.

src/redux/actions/concepts/dictionaryConcepts.js
```diff
@@ -436,9 +434,8 @@ export const fetchExistingConcept = conceptUrl => async (dispatch) => {
...
+    if (error && error.response) {
```
Karuhanga: Missed one here :)

judeinno: Done please review @Karuhanga

PR #559, opened 2019-09-04
OCLOMRS-714: Move important sources to top of list when selecting sources on the in the create concept form

judeinno:
# JIRA TICKET NAME:
[Move important sources to top of list when selecting sources on the in the create concept form](https://issues.openmrs.org/browse/OCLOMRS-714)

# Summary:
When creating a new request the sources should be displayed in order of priority as stated  [here](https://talk.openmrs.org/t/developing-the-ocl-for-openmrs-application/17771/1230?u=judeatu)

### DIFF
--- a/src/helperFunctions.js
+++ b/src/helperFunctions.js
@@ -11,3 +11,45 @@ export const checkErrorMessage = (error, message) => {
     notify.show(message, 'error', 3000);
   }
 };
+export const prioritiseItems = (arrayToSort, arrayToOrderBy, keyToSortBy) => {
+  arrayToSort.sort((a, b) => {
+    const A = a[keyToSortBy];
+    const B = b[keyToSortBy];
+
+    if (arrayToOrderBy.indexOf(A) > arrayToOrderBy.indexOf(B)) {
+      return -1;
+    }
+    return 1;
+  });
+
+  return arrayToSort;
+};
+
+export const sourcePriorityOrder = [
+  'CIEL',
+  'SNOMED CT',
+  'ICD-10-WHO',
+  'ICD-10-WHO 2nd',
+  'LOINC',
+  'RxNORM',
+  'IMO ProblemIT',
+  'IMO ProcedureIT',
+  '3BT',
+  'ICPC2',
+  'AMPATH',
+  'PIH',
+  'PIH Malawi',
+  'org.openmrs.module.mdrtb',
+  'CCAM',
+  'HL-7 CVX',
+  'NDF-RT NUI',
+  'MED-RT NUI',
+  'NCI Concept Code',
+  'FDA Route of Administration',
+  'org.openmrs.module.emrapi',
+  'HL7 DiagnosticServiceSections',
+  'HL7 2.x Route of Administration',
+  'HL7 DiagnosticReportStatus',
+  'SNOMED US',
+  'Radlex',
+];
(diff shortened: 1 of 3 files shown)


src/helperFunctions.js
```diff
@@ -11,3 +11,45 @@ export const checkErrorMessage = (error, message) => {
     notify.show(message, 'error', 3000);
   }
 };
+export const mapOrder = (arrayToSort, arrayToOrderBy, keyToSortBy) => {
+  arrayToSort.sort((a, b) => {
```
Karuhanga: Does this sort method account for items that might not be in the list?

src/helperFunctions.js
```diff
@@ -11,3 +11,45 @@ export const checkErrorMessage = (error, message) => {
     notify.show(message, 'error', 3000);
   }
 };
+export const mapOrder = (arrayToSort, arrayToOrderBy, keyToSortBy) => {
```
Karuhanga: I think this name could be more descriptive. Something like `sortSources` maybe?.

src/helperFunctions.js
```diff
@@ -11,3 +11,45 @@ export const checkErrorMessage = (error, message) => {
...
+    if (arrayToOrderBy.indexOf(A) > arrayToOrderBy.indexOf(B)) {
+      return 1;
+    }
+    return -1;
+  });
+
+  return arrayToSort;
+};
+
+export const sourcePriorityOrder = [
```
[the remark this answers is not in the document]
judeinno: Yes in the response returned from the backend the names are returned as IDs [here](https://github.com/openmrs/openmrs-ocl-client/pull/559/files#diff-e812e5fb3d2d789b04b88c4cfc2aefb5R81)

src/helperFunctions.js
judeinno: I thought this being a helper method that could be used somewhere else it needed to have a general name. Just in case in the future, we need to map another array against a specific order we could use the same function.

src/helperFunctions.js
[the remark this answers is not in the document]
judeinno: Before I run the function when the sources are returned, I add them to the end of the list  [here](https://github.com/openmrs/openmrs-ocl-client/pull/559/files#diff-e812e5fb3d2d789b04b88c4cfc2aefb5R84) so that the sources are sorted according to the list provided by Andrew [here](https://talk.openmrs.org/t/developing-the-ocl-for-openmrs-application/17771/1230?u=judeatu) then the rest that may not be in that list come last. And since this is a helper function that could be used again I didn't find it wise to add them directly here.

src/helperFunctions.js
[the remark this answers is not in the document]
Karuhanga: Some of these are names an not `id`s. `PIH Malawi` for example. Here(https://api.demo.openconceptlab.org/sources/?verbose=true&q=PIH) the id for that is `PIH-Malawi`

src/helperFunctions.js
Karuhanga: That makes sense, but I still think the name could improve. `mapOrder` is still not very clear. `prioritiseItems` maybe?

src/helperFunctions.js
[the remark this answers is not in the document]
Karuhanga: I see. But if this is indeed a helper function, can we always assume that whoever uses it in the future will use a list which has all the expected items? I think it would be safer to account for the situation where both `indexOf`s return -1.

src/helperFunctions.js
judeinno: Alright, let me work on it.

dkayiwa:
@judeinno did you see the merge conflicts

PR #568, opened 2020-01-14
OCLOMRS-737:Changed text from Add set... to Add set member

HerbertYiga:
Changed text from Add set... to Add set member

Ticket: https://issues.openmrs.org/browse/OCLOMRS-737


HerbertYiga:
Hi lincon,what could cause be causing the failing tests here @Karuhanga

PR #616, opened 2020-09-30
OCLOMRS-878:Previous Concept Form Validation Error Message Persists into New Custom Concept Form

jwnasambu:
# JIRA TICKET NAME:
https://issues.openmrs.org/browse/OCLOMRS-878

# Summary:

I added a return function to the effect. This function is to be  called when the effect needs to be cleaned up.

### DIFF
--- a/src/apps/concepts/pages/CreateOrEditConceptPage.tsx
+++ b/src/apps/concepts/pages/CreateOrEditConceptPage.tsx
@@ -113,12 +113,14 @@ const CreateOrEditConceptPage: React.FC<Props> = ({
   useEffect(() => {
     // only retrieve the concept if the context was edit at the beginning
     // otherwise we obviously have nothing to edit
-    if (originallyEditing) retrieveConcept(conceptUrl);
-    retrieveDictionary(linkedDictionaryUrl);
+    return () => { resetConceptForm(); }
 
     // usually doing the following is a mistake and will bite us later
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, []);
+  useEffect(() => {
+    return () => { resetConceptForm(); }
+}, []);
 
   // everything went hunky-dory, and we should redirect the user to the view concept page
   if (!loading && previouslyLoading && concept && !errors && !anyMappingsErrors)


kirity-mahanty:
@jwnasambu Can you please look into the build pipeline.

jwnasambu:
@kirity-mahanty  thanks for the review but am still working on it.

src/apps/concepts/pages/CreateOrEditConceptPage.tsx
```diff
@@ -113,12 +113,14 @@ const CreateOrEditConceptPage: React.FC<Props> = ({
   useEffect(() => {
     // only retrieve the concept if the context was edit at the beginning
     // otherwise we obviously have nothing to edit
-    if (originallyEditing) retrieveConcept(conceptUrl);
-    retrieveDictionary(linkedDictionaryUrl);
+    return () => { resetConceptForm(); }
```
ibacher, line 116: We actually don't want to touch *this* `useEffect()` call or at least we don't want to overwrite it. I suppose what we should have here is this:

```js
 useEffect(() => {
    // only retrieve the concept if the context was edit at the beginning
    // otherwise we obviously have nothing to edit
    if (originallyEditing) retrieveConcept(conceptUrl);
    retrieveDictionary(linkedDictionaryUrl);

    return () => resetConceptForm();

    // usually doing the following is a mistake and will bite us later
}, []);
```

This would replace everything in to commit, as we only need to return the function that calls `resetConceptForm()` once.

ibacher:
Adding to my comment, now we actually need to start defining the `resetConceptForm` function. This needs to be done in a few places. First, because we're using Typescript (a superset of Javascript that adds typing information) we need to actually add the relevant type information for `resetConceptForm`.

To do this, find the `ActionProps` interface. This interface defines the various functions that are passed as properties to the `CreateOrEditConceptPage` object. You'll see there are 3 function interfaces defined there which specify the kinds of arguments the functions take and their return values. For our `resetConceptForm` function, things are pretty simple since we don't take any arguments and don't return any value, so it's signature looks like this:

```ts
resetConceptForm: () => void;
```

Then we need to actually import the `resetConceptForm` function into the scope of the `CreateOrEditConceptPage` object. To do that, we just need to add `resetConceptForm` to the list of properties somewhere, possibly at the beginning or the end of the list of properties (this is for the list of parameters starting around line 71).

Finally, near the end of the `CreateOrEditConceptPage` file there's an object called `mapActionsToProps` that contains the definitions that map between actions (functions that modify the Redux store) and properties. To this we need to add an entry for the `resetConceptForm` property that looks something like this:

```ts
resetConceptForm: resetConceptFormAction
```

PR #627, opened 2020-10-15
OCLOMRS-889: Don't pull in concepts from other sources for mappings

jwnasambu:
# JIRA TICKET NAME:
https://issues.openmrs.org/browse/OCLOMRS-889

# Summary:
I removed the [mappings, MAPPINGS_BATCH_INDEX, "mappings"] element so that we no longer try to add a concept to the dictionary simply because we have mapped one of our concepts to it.

### DIFF
--- a/src/apps/concepts/redux/actions.ts
+++ b/src/apps/concepts/redux/actions.ts
@@ -174,8 +174,11 @@ export const upsertConceptAndMappingsAction = (
       );
 
       const state: AppState = getState();
+      // we *only* want to import either concept answers or members of the concept set
       const toConceptUrls: string[] = [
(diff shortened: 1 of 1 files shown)


ibacher:
@jwnasambu After some quick testing, it appears I led you astray on where this needed to be fixed. I've updated the version in your PR to reflect the necessary changes. To get the update locally, you probably need to remove the OCLOMRS-889 branch like so:

```
git checkout master
git branch -D OCLOMRS-889
git fetch --all --prune
git checkout OCLOMRS-889
```

jwnasambu:
@ibacher  thanks for the review and testing. I have seen the updates.

PR #631, opened 2020-10-19
Vinisha, Kirity | OCLOMRS-895 | MOBN-1475 | Clear status state and auth state on logout

vinishad:
# JIRA TICKET NAME:
[OCLOMRS-895](https://issues.openmrs.org/browse/OCLOMRS-895)

# Summary:
Even after logout few things are not cleared form state while is resulting in some improper behaviour in notifications page.

In some of the scenarios, in notification bar user getting long error message for failed import. We need to see if we can reduce the error message to get a comfortable view for user.


HerbertYiga:
Thanks @vinishad  for this, am seeing react is not recognizing the camel case this can be seen in the ci logs

PR #641, opened 2020-11-26
OCLOMRS-874:Concept selections in "Pick from existing" disappear if user extends list length

jwnasambu:
# JIRA TICKET NAME:
https://issues.openmrs.org/browse/OCLOMRS-874

# Summary:
I deleted the
```
 useEffect(() => {
    resetSelected();
``` 
Which was  resetting the  selected concepts when new ones are fetched.


ibacher (commented):
@jwnasambu I like this fix! It's very simple and to-the-point. One quick thing: we don't want to just leave a commented-out block of code, so just delete all the code from lines 195-202 (that's the `useEffect()` block.

jwnasambu:
@ibacher I have fixed the proposed changes hope they meet the ticket requirement. Thanks for reviewing my work.

PR #647, opened 2021-01-08
OCLOMRS-907:Sort Names Alphabetically by Language (so Long & Short names end up grouped together)

jwnasambu:
# JIRA TICKET NAME:
[https://issues.openmrs.org/browse/OCLOMRS-907)

# Summary:

### DIFF
--- a/src/apps/concepts/components/ConceptForm.tsx
+++ b/src/apps/concepts/components/ConceptForm.tsx
@@ -122,8 +122,17 @@ const buildInitialValues = (
 ) => ({
   ...initialValues,
   concept_class: conceptClass,
-  names
-});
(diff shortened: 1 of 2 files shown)


ibacher:
@jwnasambu Thanks for this work! We actually need to make the `buildInitialValues()` function into more of a function like this:

```ts
const buildInitialValues = (
  conceptClass: string = initialValues.concept_class,
  names: ConceptName[]
) => {
  // sort the names
  names.sort((n1: ConceptName, n2: ConceptName) => {
    if (n1.locale !== n2.locale) {
      return n1.locale < n2.locale ? -1 : 1;
    }

    if (n1.name !== n2.name) {
      return n1.name < n2.name ? -1 : 1;
    }

    return 0;
  });

  return {
    ...initialValues,
    concept_class: conceptClass,
    names
  };
}
```

PR #648, opened 2021-01-08
OCLOMRS-809: Implement viewing, editing, and deleting organisation.

hadijahkyampeire:
# JIRA TICKET NAME:
[OCLOMRS-809](https://issues.openmrs.org/browse/OCLOMRS-809)

# Summary:
- Other tickets are editing, deleting, auto-refresh, moving the orgs icon down, and adding a toast;
- [OCLOMRS-903](https://issues.openmrs.org/browse/OCLOMRS-903)
- [OCLOMRS-915](https://issues.openmrs.org/browse/OCLOMRS-915)
- [OCLOMRS-908](https://issues.openmrs.org/browse/OCLOMRS-908)
- [OCLOMRS-909](https://issues.openmrs.org/browse/OCLOMRS-909)
- [OCLOMRS-910](https://issues.openmrs.org/browse/OCLOMRS-910)
- Besides that, I made the cards for dictionaries and sources clickable.
- Also made some tweaks to the search query updating the search input
- Also added pagination to organizations.


src/apps/containers/components/EditMenu.tsx
```diff
@@ -30,6 +39,13 @@ export const EditMenu: React.FC<Props> = ({backUrl}: Props) => {
                         Discard changes and view
                     </Link>
                 </MenuItem>
+                {hasDelete && 
+                <MenuItem>
+                    <Button onClick={openModal} className={classes.deleteButton}>
+                        Delete Organisation
```
ibacher, line 45: The text "Delete Organisation" shouldn't be embedded in a component outside of the organisation app. Similarly, the way this renders in my browser seems off:

Screenshot 2021-01-11 at 15 55 43

I.e., the text is smaller and there's some additional padding. I would think that we should make this look more like the "Retire Concept" action when editing a concept. I also don't think we need to make the text red, but that's just my opinion.

src/apps/containers/components/EditMenu.tsx
```diff
@@ -1,15 +1,24 @@
...
+import {Button, Fab, Menu, MenuItem, Tooltip, makeStyles} from "@material-ui/core";
 import {MoreVert as MenuIcon} from "@material-ui/icons";
 import {Link} from "react-router-dom";
 import React from "react";
 import {useAnchor} from "../../../utils";
 
 interface Props {
     backUrl: string;
+    hasDelete?: boolean;
+    openModal?: () => void;
```
ibacher, line 10: I think this could be named better, e.g., `confirmDelete` or `handleDelete`. Something that makes it clear what it's tied to.

src/apps/organisations/api.ts
```diff
@@ -1,18 +1,39 @@
 import { authenticatedInstance } from "../../api";
-import { Organisation } from './types';
+import { Organisation, EditableOrganisationFields } from './types';
 import { AxiosResponse } from "axios";
 
 const api = {
   getUserOrgs: (username:string) => authenticatedInstance.get(`/users/${username}/orgs/?sortDesc=lastUpdated`),
-  retrieve: (username:string, q: string = "") => authenticatedInstance.get(`/users/${username}/orgs/?limit=0`, {
+  retrieve: {
```
ibacher, line 7: What is this supposed to retrieve? All orgs? The orgs for a specific user? This should probably be modelled more closely to the apis for dictionaries and sources.

ibacher (changes requested):
@hadijahkyampeire First of all, thank you for all your hard work on this! This looks to be a very valuable addition to the OCL for OpenMRS app. Since this is a very large PR, it's hard to review it all in one go, so I've added some quick comments on the stuff I managed to get through in a hour.

At a high level, one problem I see is that *when I'm logged in as `admin`*, I can access pretty much any org by going to the page `/user/orgs/:org` (this is great!). However, when I'm logged in as a normal user who is not a member of any org, going to `/user/orgs/:org` is both a bit nonsensical and doesn't work. Moreover, what I'd expect to be the "main page" to access an organisation (e.g. /orgs/:org) doesn't actually work. So, e.g., going to `/orgs/CIEL` renders... an empty screen.

Again, thanks for all your hard work on this and I hope this review isn't discouraging!

hadijahkyampeire:
Thanks @ibacher for the review, I will work on it first thing tomorrow, it's really helpful.

src/apps/containers/components/EditMenu.tsx
gracepotma: Agreed with Ian's comments; this text should be the same size, style, and color as the "Discard changes and view" option. We can sleep in peace though because we know that if a user accidentally clicks "Delete Organization", they will see the confirm modal. Right?

ibacher:
@hadijahkyampeire Please ping me when you get the code changes you demonstrated today into this branch. I'm really excited to have this done!

hadijahkyampeire:
@ibacher the internet is not yet fully restored, I am not able to push the changes I have them committed waiting to push. I will definitely ping you when I update.

hadijahkyampeire:
@ibacher I have pushed the updates and PR feedback.

src/apps/organisations/redux/actions.ts
```diff
@@ -4,19 +4,48 @@ import {
...
-import { Organisation } from '../types';
+import {PERSONAL_ORGS_ACTION_INDEX} from "./constants";
+import { Organisation, EditableOrganisationFields } from '../types';
 
 const createOrgsAction = createActionThunk(
   CREATE_ORGANISATION_ACTION,
   api.create
 );
 
+const editOrgsAction = createActionThunk(
```
ibacher, line 27: One thing I don't really understand here is why we have both this `editOrgsAction` and `editOrganisationAction` below. Do we actually need these? (I'd tend to prefer the versions using `createActionThunk` because that hooks in our general error handling).

ibacher (approved):
@hadijahkyampeire I think this mostly looks ok. One question for you below.

src/apps/organisations/redux/actions.ts
hadijahkyampeire: I guess I made a repetition there, I will clear that.

PR #651, opened 2021-01-19
OCLOMRS-918:Added a AddConceptIcon for adding a Public Concept to my …

nk183:
OCLOMRS-918:Add a Public Concept to my dictionary from the details view

[OCLOMRS-918](https://issues.openmrs.org/browse/OCLOMRS-918)

# Summary:
Added an "AddConceptToDictionaryIcon" in "Detailed View of concept" for adding the Public Concept to my Dictionary

Firstly Imported "addConceptToDictionary function" and a boolean value which tells whether to show the button(AddConceptToDictionaryIcon) or not , to the "ViewConceptPage" 
Then called the component "AddConceptToDictionaryIcon" giving argument, "addConceptToDictionary function" ,"canAddConcept" boolean and the Concept to add.

image


HerbertYiga:
@nk183  have you seen some breaking test as reported by cypress in your ci logs

nk183:
@HerbertYiga  now I think everything is fine

nk183:
@HerbertYiga @ibacher please review this PR

src/apps/concepts/components/AddConceptToDictionaryIcon.tsx
```diff
@@ -0,0 +1,43 @@
+import React from "react";
+import {
+    Fab,
+    Tooltip
+} from "@material-ui/core";
+import {APIConcept} from "../types";
+
```
ibacher, line 7: Could you clean-up the blank lines and the spacing / indentation a bit? (We tend to use 2 spaces to indent on this project)

src/apps/concepts/components/AddConceptToDictionaryIcon.tsx
```diff
@@ -0,0 +1,43 @@
...
+import {Add as AddIcon} from "@material-ui/icons";
+
+
+
+interface AddConceptToDictionaryIconProps {
+    dictionaryToAddTo:string;
+    canAddConcept:string;
```
ibacher, line 14: Shouldn't this be a boolean?

src/apps/concepts/components/AddConceptToDictionaryIcon.tsx
```diff
@@ -0,0 +1,43 @@
...
+    row:APIConcept|undefined ;
```
ibacher, line 15:
```suggestion
    row?: APIConcept ;
```

ibacher: I think this could do with a better name...

src/apps/concepts/components/AddConceptToDictionaryIcon.tsx
```diff
@@ -0,0 +1,43 @@
...
+    addConceptsToDictionary: Function;
```
ibacher, line 16:
```suggestion
    addConceptsToDictionary: (APIConcept[]) => void;
```

src/apps/concepts/components/AddConceptToDictionaryIcon.tsx
```diff
@@ -0,0 +1,43 @@
...
+          dictionaryToAddTo,
+          canAddConcept,
+          row,
+          addConceptsToDictionary
+        } = props;
+    return (canAddConcept=="false"  || dictionaryToAddTo==null || dictionaryToAddTo==undefined)? (null):
+        (
+            <Tooltip title="Add concept to Dictionary" data-testid="addConceptToDictionaryIcon">
+                <Fab onClick={() => {
+                                if (addConceptsToDictionary){
```
ibacher, line 32: Do we need this `if` check?

src/apps/concepts/components/AddConceptToDictionaryIcon.tsx
```diff
@@ -0,0 +1,43 @@
...
+          row,
+          addConceptsToDictionary
+        } = props;
+    return (canAddConcept=="false"  || dictionaryToAddTo==null || dictionaryToAddTo==undefined)? (null):
+        (
+            <Tooltip title="Add concept to Dictionary" data-testid="addConceptToDictionaryIcon">
+                <Fab onClick={() => {
+                                if (addConceptsToDictionary){ 
+                                    addConceptsToDictionary([row]);
+                                    window.history.back() ;
```
ibacher, line 34: Why are we changing the page the user is on?

src/apps/concepts/components/AddConceptToDictionaryIcon.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+
+    export function AddConceptToDictionaryIcon(props: AddConceptToDictionaryIconProps) {
+        const {
+          dictionaryToAddTo,
+          canAddConcept,
+          row,
+          addConceptsToDictionary
+        } = props;
+    return (canAddConcept=="false"  || dictionaryToAddTo==null || dictionaryToAddTo==undefined)? (null):
```
ibacher, line 28: Please use `===` instead of `==`. Also, at least per the typing, `dictionaryToAddTo` cannot possibly be `null` or `undefined` at this point. Last point on this: if you're going to check if something is `null` or `undefined`, check the `undefined` first.

```ts
return (!canAddConcept || dictionaryToAddTo === undefined || dicationaryToAddTo === null) ? null : ( ...
```

src/apps/concepts/components/ConceptsTableRow.tsx
```diff
@@ -90,7 +92,7 @@ const conceptNameCell = (
     >
       <Link
         onClick={(e) => e.stopPropagation()}
-        to={`${row.version_url}?linkedDictionary=${linkedDictionary}`}
+        to={`${row.version_url}?linkedDictionary=${linkedDictionary}&dictionaryToAddTo=${dictionaryToAddTo}&canAddConcept=${buttons.addToDictionary}`}
```
ibacher, line 95: It's bad to populate a security concept (`canAddConcept` via a URL parameter)

src/apps/concepts/components/AddConceptToDictionaryIcon.tsx
nk183: yes but as a query string Parameter it is taking 'String' as the default data type

src/apps/concepts/components/AddConceptToDictionaryIcon.tsx
nk183: `window.history.back() ;`
will definitely direct the user on "Import existing concept from CIEL" (ViewConceptPage)
and it is required in  Issue Discription

src/apps/concepts/components/AddConceptToDictionaryIcon.tsx
nk183: thank you @ibacher this was a blunder
`dictionaryToAddTo cannot possibly be null or undefined at this point.`
it can be this is why the previous submission was [failing to build The Travis CI](https://talk.openmrs.org/t/i-dont-know-how-to-tackle-the-travis-ci-build-failed/31763).

src/apps/concepts/components/AddConceptToDictionaryIcon.tsx
ibacher: So, you want think about this flow a little more carefully. What happens when the user lands directly on this page instead of by navigating from the app? This can happen if, for example they copy and paste the link to this page into the browser. Calling `window.history.back()` moves them back in history, but not necessarily where we want them to be in the application. It may be helpful to look at how things like this are handled in the app at present, i.e., when creating a new concept, I can be redirected back to the page containing the concepts in the dictionary.

src/apps/concepts/components/AddConceptToDictionaryIcon.tsx
ibacher: Then we need to change the typing of `dictionaryToAddTo` to something like `dictionaryToAddTo?: string` to reflect that.

src/apps/concepts/components/ConceptsTableRow.tsx
nk183: @ibacher  please any suggestion for alternate to that 
how can I pass the `canAddConcept` value

ibacher: You'll basically need to populate this the same way we do [when determining whether a user can modify a dictionary](https://github.com/openmrs/openmrs-ocl-client/blob/a0df00455fbae8484244366e4420389599a80f80/src/apps/concepts/pages/ViewConceptsPage.tsx#L217), i.e., by calling `canModifyContainer()`. `ownerType` and `owner` probably need to be derived from the `dictionaryToAddTo` parameter

nk183: hi, @ibacher I tried to do the above , But It is not working.
If you say can we encrypt the `canAddConcept` and pass in the same way (as a query parameter).

src/apps/concepts/components/AddConceptToDictionaryIcon.tsx
nk183: @ibacher why we do `addConceptsToDictionary: (APIConcept[]) => void;`

src/apps/concepts/components/ConceptsTableRow.tsx
nk183: hi @ibacher, @HerbertYiga,  please tell.

src/apps/concepts/components/AddConceptToDictionaryIcon.tsx
nk183: @ibacher see [here](https://github.com/openmrs/openmrs-ocl-client/blob/2dcb3ceebc21dea96044e5eddd7c333bd506bf6d/src/apps/concepts/components/ConceptsActionMenu.tsx#L27)
it is I tried to replicate it

nk183:
@ibacher I have tried to fix some bugs.
please tell I should do now about this issue.

nk183:
@ibacher  @HerbertYiga @hadijahkyampeire, please review my PR

src/apps/concepts/components/AddConceptToDictionaryIcon.tsx
nk183: but @ibacher  as default datatype of query param, it is String

src/apps/concepts/components/AddConceptToDictionaryIcon.tsx
ibacher: Better typing. `Function` is quite generic; `(APIConcept[]) => void` allows me to be certain that anything we're plugging into `addConceptsToDictionary` is a function that takes an array of `APIConcept` objects and returns void. In turn this greatly simplifies the required function.

ibacher:
@nk183 Please see [this commit](https://github.com/openmrs/openmrs-ocl-client/pull/651/commits/75573121b7f9ea474083944d60f27d91056e5f2a) I've added to this branch that implements the SpeedDial component I alluded to on the call and gets rid of the the `canAddConcept` URL parameter.

ibacher (approved):

ibacher:
Merged in [this commit](https://github.com/openmrs/openmrs-ocl-client/commit/6a6286ff348e7e1970952c547eabe87d2af58e1d).

PR #655, opened 2021-01-26
OCLOMRS-890:added clear error message for :release ID already exists

nk183:
# JIRA TICKET NAME:
[OCLOMRS-890:](https://issues.openmrs.org/browse/OCLOMRS-890)

# Summary:
image
image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8516,9 +8516,9 @@
       "integrity": "sha512-k/vGaX4/Yla3WzyMCvTQOXYeIHvqOKtnqBduzTHpzpQZzAskKMhZ2K+EnBiSM9zGSoIFeMpXKxa4dYeZIQqewQ=="
     },
     "ini": {
-      "version": "1.3.5",
-      "resolved": "https://registry.npmjs.org/ini/-/ini-1.3.5.tgz",
-      "integrity": "sha512-RZY5huIKCMRWDUqZlEi72f/lmXKMvuszcMBduliQ3nnWbx9X/ZBQO7DijMEYS9EhHBb2qacRUMtC7svLwe0lcw=="
+      "version": "1.3.8",
+      "resolved": "https://registry.npmjs.org/ini/-/ini-1.3.8.tgz",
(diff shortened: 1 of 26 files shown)


nk183:
please help 
it working fine on my pc and don't know why it is failing checks

HerbertYiga:
@k-joseph  @dkayiwa  do you have any thoughts on the cypress  error for this!!

ibacher (changes requested):
@nk183 Thanks for this and for your work digging out the cause! That said, I think we should take a slightly different approach which is to modify the `errorMsgResponse()` function so that:

1. If there's a "detail" element in the response data, we just return the detail
2. If there's not a "detail" element in the response data, we return the response the way we do today

nk183:
>     2. If there's not a "detail" element in the response data, we return the response the way we do today

I am not able to understand by `the response the way we do today`

ibacher:
@nk183 So, I'm suggesting modifying the [`errorMessageResponse()`](https://github.com/openmrs/openmrs-ocl-client/blob/a0df00455fbae8484244366e4420389599a80f80/src/redux/utils.ts#L103) function to some thing like:

```ts
export function errorMsgResponse(response: any) {
  let errorMsgResponse = [];
  const genericErrorMessage =
      "Action could not be completed. Please retry.";

  if (response.data && response.data.has('details') {
    errorMsgResponse.push(response.data.details);
  } else {
    for (let key in response.data) {
      errorMsgResponse.push(
            ? response.data[key].join(',')
            : response.data[key]
      )
    };
  }
  return errorMsgResponse.length > 0 ? errorMsgResponse.join('\n') : genericErrorMessage;
};
```

nk183:
`for (let key in response.data) {
      errorMsgResponse.push(
            ? response.data[key].join(',')
            : response.data[key]
      )
 };`
@ibacher why we are combining all "key value"?

nk183:
@ibacher Can we just do this::
`export function errorMsgResponse(response: any) {
  let errorMsgResponse = [];
  const genericErrorMessage =
      {detail:"Action could not be completed. Please retry."};
  if(response.data){
    return response.data;
  }
  return genericErrorMessage;
};
`

hadijahkyampeire:
@nk183 I think we should go with @ibacher 's function, that way we are sure we won't break anything because I am sure by the time they made that helper, they were considering various formats of errors and wanting to return them in the best way that is easy to work with, in our components.
So let's just add in the case for `details` since it's what we need now.

PR #656, opened 2021-01-28
OCLOMRS-924: Adds new member to an organisation.

hadijahkyampeire:
# JIRA TICKET NAME:
[Add new member to an organization](https://issues.openmrs.org/browse/OCLOMRS-924)

# Summary:

# Caveats:
- I can't test it working yet, not sure if it's because of permissions, but I am reaching out to the backend people to help me.

### DIFF
--- a/src/apps/organisations/redux/actions.ts
+++ b/src/apps/organisations/redux/actions.ts
@@ -126,6 +126,11 @@ const retrieveOrgMembersAction = createActionThunk(
     api.organisation.retrieveMembers
 );
 
+const addOrgMemberAction = createActionThunk(
+  GET_ORG_MEMBERS_ACTION,
+  api.organisation.AddMember
+);
+
 export { 
   createOrganisationAction,
   resetCreateOrganisationAction,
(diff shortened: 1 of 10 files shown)


ibacher:
@hadijahkyampeire I've made some adjustments to this, but this works for me (I had to create a new organisation to get it to work). A couple of things I think we need to take account of:

1. The modal form doesn't disappear when it's submitted whether successfully or unsuccessfully.
2. In the case where we get a 403 error (or, indeed, another error) we should give the user some kind of feedback (maybe a notification in the notification bit?)

hadijahkyampeire:
Thanks @ibacher , I am going to work on the notification and modal closing.

src/apps/organisations/redux/actions.ts
```diff
@@ -122,13 +123,13 @@ const deleteOrganisationAction = createActionThunk(
...
-    GET_ORG_MEMBERS_ACTION,
-    api.organisation.retrieveMembers
+  GET_ORG_MEMBERS_ACTION,
+  api.organisation.retrieveMembers
 );
 
 const addOrgMemberAction = createActionThunk(
-  GET_ORG_MEMBERS_ACTION,
-  api.organisation.AddMember
+  CREATE_ORG_MEMBER_ACTION,
```
hadijahkyampeire, line 131: Sorry about this @ibacher , how did I miss that, it's my bad.

hadijahkyampeire:
@ibacher , I catch the error message and I am displaying it in an Alert notification, but I couldn't test the successful adding, hope it works. Also, I close the dialog when loading is false, so that should cater for success or failure.

PR #659, opened 2021-01-31
OCLOMRS-927: Export head dictionary as CSV and JSON

hadijahkyampeire:
# JIRA TICKET NAME:
[Bulk Export of dictionaries to spreadsheet and JSON](https://issues.openmrs.org/browse/OCLOMRS-901)

# Summary:
[OCLOMRS-972](https://issues.openmrs.org/browse/OCLOMRS-927)

# Caveats
- I used another library `export-from-json` because it can easily export data in most forms hence `reusable` and deals well with nested data. cc @ibacher

### DIFF
--- a/src/apps/dictionaries/components/DictionaryConceptsSummary.tsx
+++ b/src/apps/dictionaries/components/DictionaryConceptsSummary.tsx
@@ -1,4 +1,5 @@
 import React from "react";
+import exportFromJSON from 'export-from-json'; //This seems to be reusuable in all sorts of exports i.e json,csv,txt etc
 import { APIDictionary } from "../../dictionaries";
 import {
   Button,
@@ -8,12 +9,11 @@ import {
   Typography,
   Menu,
   MenuItem,
-  IconButton,
+  IconButton
 } from "@material-ui/core";
 import { MoreVert, TableChartOutlined, FormatListBulleted } from "@material-ui/icons";
 import { Link } from "react-router-dom";
-import {useAnchor} from "../../../utils";
-import { CSVLink } from "react-csv";
+import {useAnchor, getDate} from "../../../utils";
 
 interface Props {
   dictionary: APIDictionary;
@@ -55,13 +55,10 @@ const DictionaryConceptsSummary: React.FC<Props> = ({ dictionary }) => {
   ).length;
   const customConceptCount = conceptReferences.length - fromPreferredSource;
 
-  const csvHeaders = [
-    { label: 'Concept Expression', key: 'expression' },
-    { label: 'Reference Type', key: 'reference_type' }
-  ];
 
   const { updated_on = ''} = dictionary || {};
-  // console.log(new Date(updated_on).toIsoDate())
+  const updatedDate = getDate(updated_on);
+
   return (
     <Paper className="fieldsetParent">
       <fieldset>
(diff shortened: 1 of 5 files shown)


src/apps/dictionaries/components/DictionaryConceptsSummary.tsx
```diff
@@ -36,13 +55,40 @@ const DictionaryConceptsSummary: React.FC<Props> = ({ dictionary }) => {
   ).length;
   const customConceptCount = conceptReferences.length - fromPreferredSource;
 
+
+  const { updated_on = ''} = dictionary || {};
+  const updatedDate = getDate(updated_on);
```
ibacher, line 60:
```suggestion
  const updatedDate = dictionary.updated_on;
```

src/utils/utils.ts
```diff
@@ -115,4 +115,12 @@ export function stableSort<T>(array: T[], comparator: (a: T, b: T) => number) {
...
\ No newline at end of file
+}
+
+export const getDate = (date: any) => {
+  if (typeof date === 'object') {
+    const isoDate = date.toISOString();
+    return isoDate.substr(0, isoDate.indexOf('T'));
+  }
+  return typeof date === 'string' ? date.substr(0, date.indexOf('T')) : date;
+};
```
ibacher, line 126:
```suggestion
```

src/apps/dictionaries/components/DictionaryConceptsSummary.tsx
ibacher: We also need to update the the `APIDictionary` type to have the `updated_on` property as a `string` property. One of the real advantages of Typescript, used properly, is that we can remove ambiguity around what type of data we're handling. In this case, `dictionary` is an instance of `APIDictionary` and it will have the `updated_on` property (that's guaranteed by the API) with a `string` value.

I'm also suggesting we don't reformat the string returned from the API. The reason is that the HEAD version of a dictionary can be modified multiple times on the same day, so we should just use the updated time stamp to determine the export version. It *might* be worth looking into what happens when a dictionary has just been created but never had any concepts added to it. In that instance `updated_on` might be `undefined` and we should fall back to `created_on`, which would mean adding *that* property to `APIDictionary` and changing this line to:

```ts
const updatedDate = dictionary.updated_on || dictionary.created_on;
```

src/apps/dictionaries/components/DictionaryConceptsSummary.tsx
```diff
@@ -36,13 +55,40 @@ const DictionaryConceptsSummary: React.FC<Props> = ({ dictionary }) => {
...
+          <MoreVert />
+        </IconButton>
+        <Menu
+          id='long-menu'
+          anchorEl={menuAnchor}
+          open={Boolean(menuAnchor)}
+          onClose={handleMenuClose}>
+            <MenuItem onClick={() => exportFromJSON({data: [dictionary], fileName: `${dictionary.name}_head_${updatedDate}.csv`, exportType: 'csv'})}>
+              <TableChartOutlined /> 
+              <span className={classes.addLeftPadding}>Export Head as CSV</span>
```
ibacher, line 84:
```suggestion
              <span className={classes.addLeftPadding}>Export HEAD as CSV</span>
```

src/apps/dictionaries/components/DictionaryConceptsSummary.tsx
```diff
@@ -36,13 +55,40 @@ const DictionaryConceptsSummary: React.FC<Props> = ({ dictionary }) => {
...
+            </MenuItem>
+            <MenuItem onClick={() => exportFromJSON({data: dictionary, fileName: `${dictionary.name}_head_${updatedDate}.json`, exportType: 'json'})}>
+              <FormatListBulleted /> 
+              <span className={classes.addLeftPadding}>Export Head as JSON</span>
```
ibacher, line 88:
```suggestion
              <span className={classes.addLeftPadding}>Export HEAD as JSON</span>
```

ibacher (changes requested):
@hadijahkyampeire Thanks for this! I've made some small comments below. However, there's a bigger issue that needs to be thought-through here which is *what we want the CSV to look like*. I'm going to make some comments on the Jira ticket related to that.

hadijahkyampeire:
This is onhold I will bring it back when things are clear

PR #661, opened 2021-02-03
OCLOMRS-926:Recursively loading concepts shouldn't try to reload the concept more than once

jwnasambu:
# JIRA TICKET NAME:
[Recursively loading concepts shouldn't try to reload the concept more than once](https://issues.openmrs.org/browse/OCLOMRS-926>)

# Summary:
I fixed the tracking of concepts being fetched an sure that each concept is only loaded a single time.

### DIFF
--- a/src/apps/dictionaries/logic.ts
+++ b/src/apps/dictionaries/logic.ts
@@ -47,11 +47,18 @@ const recursivelyFetchToConcepts = async (
     `Found ${getConceptUrls(mappingsLists).length} dependent concepts to add...`
   );
 
-  for (let i = 0; i < levelsToCheck; i += 1) {
+  const loadedConcepts = new Set();
+   for (let i = 0; i < levelsToCheck; i += 1) {
     const toConceptCodes = mappingsLists[i].map(
       mapping => mapping.to_concept_code
-    );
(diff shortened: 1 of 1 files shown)


src/apps/dictionaries/logic.ts
```diff
@@ -61,7 +68,7 @@ const recursivelyFetchToConcepts = async (
     );
   }
 
-  return getConceptUrls(mappingsLists);
+  return getConceptUrls(mappingsLists);
```
HerbertYiga, line 71: do we need this change

jwnasambu: @HerbertYiga yes we need it.

HerbertYiga: > @HerbertYiga yes we need it.

i see the line you added and that you removed are similar!

hadijahkyampeire: @HerbertYiga it shows she added just space after terminating, it's technically not a change though we should mind about proper formatting. cc @jwnasambu

hadijahkyampeire (approved):
Well done @jwnasambu ! LGTM

PR #664, opened 2021-02-16
OCLOMRS-898:As a user, I want to be able to go directly to the page I  was trying to access after logging in

jwnasambu:
# JIRA TICKET NAME:
[As a user, I want to be able to go directly to the page I was trying to access after logging in.](https://issues.openmrs.org/browse/OCLOMRS-898)

# Summary:
How to allow users access their intended url instead of redirecting them back to the home page.

My routes are

### DIFF
--- a/src/apps/authentication/LoginPage.tsx
+++ b/src/apps/authentication/LoginPage.tsx
@@ -12,6 +12,8 @@ interface Props {
   login: (...args: Parameters<typeof loginAction>) => void;
   loading: boolean;
   errors?: any;
+  nextPage: any;
+  clearNextPage: any;
 }
 
 const useStyles = makeStyles({
@@ -29,7 +31,10 @@ const LoginPage: React.FC<Props> = ({
   isLoggedIn,
   login,
   loading,
-  errors 
+  errors,
+  nextPage,
+  clearNextPage
+
 }: Props) => {
   const classes = useStyles();
 
@@ -41,7 +46,13 @@ const LoginPage: React.FC<Props> = ({
     };
   }, []);
 
-  if (isLoggedIn) return <Redirect to="/" />;
+  if (isLoggedIn) {
+  if (nextPage) {
+    clearNextPage();
+    return <Redirect to={nextPage} />;
+  }
+  return <Redirect to="/" />;
+}
   else
     return (
       <Grid
--- a/src/apps/authentication/LoginPage.tsx
+++ b/src/apps/authentication/LoginPage.tsx
@@ -1,19 +1,19 @@
 import React, { useEffect } from "react";
 import { Redirect } from "react-router-dom";
+import { connect } from "react-redux";
 import { Container, Grid, makeStyles, Typography } from "@material-ui/core";
 import { Login } from "./components";
-import { connect } from "react-redux";
 import { authErrorsSelector, authLoadingSelector } from "./redux/reducer";
-import { loginAction } from "./redux";
-import { AppState } from "../../redux";
+import { clearNextPageAction, loginAction } from "./redux";
+import { AppState } from "../../redux/types"; 
 
 interface Props {
   isLoggedIn: boolean;
   login: (...args: Parameters<typeof loginAction>) => void;
   loading: boolean;
   errors?: any;
-  nextPage: any;
-  clearNextPage: any;
+  nextPage?: string;
+  clearNextPage: (...args: Parameters<typeof clearNextPageAction>) => void;
 }
 
 const useStyles = makeStyles({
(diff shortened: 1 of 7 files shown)


src/apps/authentication/components/AuthenticationRequired.tsx
```diff
@@ -46,6 +46,13 @@ const mapStateToProps = (state: any) => ({
   profile: profileSelector(state),
   profileLoading: getUserDetailsLoadingSelector(state)
 });
+const location = {
+  pathname: '/login'
```
HerbertYiga, line 50: @jwamalwa  did you miss a coma on this line we would have some thing like  pathname: '/login',

src/apps/authentication/LoginPage.tsx
```diff
@@ -79,5 +81,10 @@ const mapStateToProps = (state: any) => ({
   errors: authErrorsSelector(state)
 });
 const mapDispatchToProps = { login: loginAction };
+let { from } = this.props.location.state || { from: { pathname: '/' } } // retrieves the URL set on the PrivateRoute component above
+  if (result.payload.data.is_admin) { // checks if user is admin to redirect to admin page
+      from = '/admin';
+   } 
+   this.props.history.push(from); // this will redirect to intended page
```
ibacher, line 88: This code needs to be moved into the React component, not here.

In particular, we need to replace the line:

```ts
if (isLoggedIn) return <Redirect to="/" />
```

With something like:

```ts
if (isLoggedIn) {
  if (nextPage) {
    clearNextPage();
    return <Redirect to={nextPage} />;
  }
  return <Redirect to="/" />;
}
```

This will handle updating the location, etc. `nextPage` is clearly a property that needs to be defined and added. It can then be populated it the `AuthenticationRequired` component by changing this line:

```ts
if (!isLoggedIn) return <Redirect to="/login" />;
```

To something slightly more sophisticated, like:

```ts
const location = useLocation();

if (!isLoggedIn) {
  const currentPage = location.pathname + location.search + location.hash;
  setNextPage(currentPage);
  return <Redirect to="/login" />;
}
```

`setNextPage()` and `clearNextPage()` should be defined in `src/apps/authentication/redux/actions.ts` (as these will be actions controlling a `nextPage` property on the redux reducer).

See how far you can get with implementing those.

jwnasambu:
@HerbertYiga  and @ibacher thanks for reviewing my work. Ian, I have in-cooperated the proposed changes Kindly feel free to review my PR it at your convenient time.

jwnasambu:
mmmmmn  I have a Travis failure even when  my local server builds successfully!

src/apps/authentication/components/AuthenticationRequired.tsx
```diff
@@ -28,7 +29,13 @@ export const AuthenticationRequired: React.FC<Props> = ({
     if (isLoggedIn) getProfile();
   }, [isLoggedIn, getProfile]);
 
-  if (!isLoggedIn) return <Redirect to="/login" />;
+const location = useLocation();
+
+if (!isLoggedIn) {
+  const currentPage = location.pathname + location.search + location.hash;
+  setNextPage(currentPage);
```
ibacher, line 36:
```suggestion
  setNextPageAction(currentPage);
```

src/apps/authentication/LoginPage.tsx
```diff
@@ -6,11 +6,15 @@ import { connect } from "react-redux";
...
+
 interface Props {
   isLoggedIn: boolean;
   login: (...args: Parameters<typeof loginAction>) => void;
   loading: boolean;
   errors?: any;
+  nextPage: any;
+  clearNextPage: any;
+
 }
```
ibacher, line 18:
```suggestion

interface Props {
  isLoggedIn: boolean;
  login: (...args: Parameters<typeof loginAction>) => void;
  loading: boolean;
  errors?: any;
  nextPage: any;
  clearNextPage: any;
}
```

ibacher (commented):
@jwnasambu A couple more review comments. Also in the LoginPage.tsx, there's a `mapStateToProps` function which is where we should populate the `nextPage` value from and a `mapDispatchToProps` which is where we should populate the `clearNextPage` from.

Probably looks something like:

```ts
const mapStateToProps = (state: any) => ({
  isLoggedIn: state.auth.isLoggedIn,
  loading: authLoadingSelector(state),
  errors: authErrorsSelector(state),
  nextPage: state.auth.nextPage
});

const mapDispatchToProps = { 
  login: loginAction
  clearNextPage: clearNextPageAction
};
```

Where `clearNextPageAction` is imported much the same as `loginAction`.

ibacher (approved):

PR #671, opened 2021-03-18
OCLOMRS-491: Restyle already added concepts and add a tick icon.

hadijahkyampeire:
# JIRA TICKET NAME:
[OCLOMRS-491](https://issues.openmrs.org/browse/OCLOMRS-491)

# Summary:
- Disable the add concept button on the CIEL concepts page when a specific concept is added.
Screenshot 2021-03-18 at 20 20 44
Screenshot 2021-03-18 at 20 19 42

### DIFF
--- a/src/apps/concepts/pages/ViewConceptsPage.tsx
+++ b/src/apps/concepts/pages/ViewConceptsPage.tsx
@@ -1,4 +1,5 @@
 import React, { useEffect, useState } from "react";
+import { includes } from 'lodash';
 import { createStyles, Grid, makeStyles, Theme } from "@material-ui/core";
 import { ConceptsTable, AddConceptsIcon } from "../components";
 import { connect } from "react-redux";
@@ -76,6 +77,7 @@ export type ActionProps = {
 
 export interface OwnProps {
   containerType: string;
+  viewDictConcepts?: boolean;
 }
 
 type Props = StateProps & ActionProps & OwnProps;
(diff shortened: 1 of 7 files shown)


src/apps/concepts/pages/ViewConceptsPage.tsx
```diff
@@ -111,9 +113,13 @@ const ViewConceptsPage: React.FC<Props> = ({
   profile,
   usersOrgs,
   containerType,
+  viewDictConcepts,
   addConceptsToDictionary,
   removeConceptsFromDictionary,
 }) => {
+  const dictionaryConcepts = dictionary?.references.map(r => r.expression);
+  const modifiedConcepts = concepts?.map(c => includes(dictionaryConcepts, c.version_url) ? {...c, added: true} : {...c});
```
ibacher, line 121: I think we better to do this mapping in the `mapStateToProps` function. That way we're working with Redux instead of fighting it.

src/apps/concepts/pages/ViewConceptsPage.tsx
```diff
@@ -146,6 +152,14 @@ const ViewConceptsPage: React.FC<Props> = ({
     addToDictionary: dictionaryToAddTo,
   } = queryParams;
 
+  // This useEffect is to fetch the dictionary while on the concepts page,
+  // before when one would refresh the page the would lose the dictionary.
+  useEffect(() => {
+    if (dictionary === undefined && dictionaryToAddTo) {
+    retrieveDictionary(dictionaryToAddTo);
```
ibacher, line 159:
```suggestion
      retrieveDictionary(dictionaryToAddTo);
```

src/apps/concepts/pages/ViewConceptsPage.tsx
```diff
@@ -254,7 +268,7 @@ const ViewConceptsPage: React.FC<Props> = ({
             component='div'
           >
             <ConceptsTable
-              concepts={concepts || []}
+              concepts={viewDictConcepts ? concepts || [] : modifiedConcepts || []}
```
ibacher, line 271:
```suggestion
              concepts={(viewDictConcepts ? concepts : modifiedConcepts) ?? []}
```

ibacher (commented):
This mostly looks good; I've made a few small, but inconsequential suggestions. 

However, I think it would be a good idea to avoid merging anything until we've fixed OCLOMRS-955, OCLOMRS-956 and OCLOMRS-951, which are all related and bug in production (the idea being to keep `master` clean until we can push a new release).

hadijahkyampeire:
Thanks @ibacher for the review, I will add a tag for do not merge, so that no one merges before we solve the bugs.

PR #673, opened 2021-03-18
Oclomrs 932

jwnasambu:
# JIRA TICKET NAME:
[General Filter: Hide Retired Concepts & Add Filter option to Include Retired Concepts](https://issues.openmrs.org/browse/<OCLOMRS-932>)

# Summary:
I include Retired Concepts option in the filter menu for both the dictionary and source view, and the view the user sees when they are in the Add Existing Concepts workflow


jwnasambu:
@ibacher, @dkayiwa  below is the Travis error am getting kindly how do I go about it though locally it builds successfully and I have shared the screenshot on the ticket link.
`12.19-alpine: Pulling from library/node
ERROR: Service 'oclclient' failed to build: toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit
The command "bash ./start_local_instance.sh" exited with 1.
19.79s$ npm run test:integration`

src/apps/concepts/pages/ViewConceptsPage.tsx
```diff
@@ -192,8 +197,9 @@ const ViewConceptsPage: React.FC<Props> = ({
       sortBy: sortBy,
       dataTypeFilters: initialDataTypeFilters,
       classFilters: initialClassFilters,
+      generalFilters: initialGeneralFilters,
       sourceFilters: initialSourceFilters,
-      includeRetired: true,
+      includeRetired: false,
```
nk183, line 202: hi @jwnasambu just confirming does this `includeRetired` property is working fine now? as I am not able to see `concepts without the Retired concepts` even after  `includeRetired: false,`

HerbertYiga:
@jwnasambu  you can also look through this for some broader picture about the error

nk183:
> @ibacher, @dkayiwa below is the Travis error am getting kindly how do I go about it though locally it builds successfully and I have shared the screenshot on the ticket link.
> `12.19-alpine: Pulling from library/node ERROR: Service 'oclclient' failed to build: toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit The command "bash ./start_local_instance.sh" exited with 1. 19.79s$ npm run test:integration`

@jwnasambu I have faced it many times it works after rebuilding the Travis CI
just try restarting the build, hopefully it will pass

jwnasambu:
@nk183 thanks so so much!  rebuilding the Travic ci  has solved the issue. Once again thank you

ibacher:
@jwnasambu @nk183 I'll be honest, I generally just ignore that error; that wouldn't be a good policy if we had better test coverage, but the number of things actually tested is pretty small.

src/apps/concepts/api.ts
```diff
@@ -48,6 +49,7 @@ const api = {
             sortBy = "bestMatch",
             dataTypeFilters = [] as string[],
             classFilters = [] as string[],
+            generalFilters =[] as string[],
```
ibacher, line 52:
```suggestion
            generalFilters = [],
```

The `as string[]` isn't necessary here.

src/apps/concepts/pages/ViewConceptsPage.tsx
```diff
@@ -156,6 +157,9 @@ const ViewConceptsPage: React.FC<Props> = ({
   const [dataTypeFilters, setInitialDataTypeFilters] = useState<string[]>(
     initialDataTypeFilters
   );
+  const [generalFilters, setGeneralFilters] = useState<string[]>(
+    initialGeneralFilters
+  );
```
ibacher, line 162:
```suggestion
  const [generalFilters, setGeneralFilters] = useState(
    initialGeneralFilters
  );
```

src/utils/constants.ts
```diff
@@ -305,3 +305,7 @@ export const CONTEXT = {
   view: "view",
   edit: "edit"
 };
+export const CONCEPT_GENERAL: string[] = [
+  "Include Retired Concepts ",
+  "Include Added Concepts"
+];
```
ibacher, line 311:
```suggestion
];

```

src/apps/concepts/api.ts
```diff
@@ -59,6 +61,7 @@ const api = {
                 [sortDirection]: sortBy,
                 ...optionallyIncludeList("datatype", dataTypeFilters),
                 ...optionallyIncludeList("conceptClass", classFilters),
+                ...optionallyIncludeList("conceptgeneral", generalFilters),
```
ibacher, line 64: We actually don't need this line here.

src/utils/constants.ts
```diff
@@ -305,3 +305,7 @@ export const CONTEXT = {
   view: "view",
   edit: "edit"
 };
+export const CONCEPT_GENERAL: string[] = [
+  "Include Retired Concepts ",
+  "Include Added Concepts"
```
ibacher, line 310: I would simplify this to just "Include Retired Concepts". I think we're handling Added Concepts in a different way and, in any case, it's a different feature.

src/utils/constants.ts
```diff
@@ -305,3 +305,7 @@ export const CONTEXT = {
   view: "view",
   edit: "edit"
 };
+export const CONCEPT_GENERAL: string[] = [
+  "Include Retired Concepts ",
```
ibacher, line 309:
```suggestion
  "Include Retired Concepts",
```

src/apps/concepts/api.ts
```diff
@@ -59,6 +61,7 @@ const api = {
...
                 ...optionallyIncludeList("source", sourceFilters),
                 timestamp: new Date().getTime(),
                 includeRetired
```
ibacher, line 67: Instead, what we want is something like `includeRetired: "Include Retired" in generalFilters`

jwnasambu: @ibacher thanks for the review! let me fix the proposed changes immediately.

nk183:
Hi @jwnasambu just concern about force-push see [here](https://talk.openmrs.org/t/git-push-origin-problem/32063/4) (Brandon's Comment)

src/apps/concepts/pages/ViewConceptsPage.tsx
hadijahkyampeire: I think we can make includeRetired to depend on the generalFilters and the status of the `includeRetired` checkbox, that way, we send the right request.
` initialGeneralFilters.includes('IncludeRetired') ? Number(1) : Number(0)`

src/apps/concepts/api.ts
```diff
@@ -48,6 +49,7 @@ const api = {
             sortBy = "bestMatch",
             dataTypeFilters = [] as string[],
             classFilters = [] as string[],
+            generalFilters = [],
             sourceFilters = [] as string[],
             includeRetired = false
```
hadijahkyampeire, line 54: This will now be changing according to the `includeRetired` checkbox filter.

src/apps/concepts/api.ts
```diff
@@ -35,6 +35,7 @@ const api = {
             sortBy?: string;
             dataTypeFilters?: string[];
             classFilters?: string[];
+            generalFilters?: string[];
             sourceFilters?: string[];
             includeRetired?: boolean;
```
hadijahkyampeire, line 40: This would then be of type number. `includeRetired?: number`

hadijahkyampeire (changes requested):
Thanks, @jwnasambu for the work, however, it is not working as expected at least making the right request with a changing param of `includeRetired`, 🤔  so I have added suggestions that can point to the right direction of this.

If somethings are not clear, we can sync about them tomorrow.  Look at this doc too, so that you see the params to pass to the URL https://github.com/OpenConceptLab/oclapi/wiki/concepts#list-concepts-in-a-source.

hadijahkyampeire:
@ibacher I might have understood this wrongly or right, could you please let us know what is expected of this PR, thanks.

src/apps/concepts/api.ts
jwnasambu: @hadijahkyampeire thanks for the review! concerning `Also my understanding is that what you are calling generalFilters, should instead be includeRetired, which becomes true and false on checking and unchecking the checkbox respectively.`  kindly look at the ticket requirement critically its not the way you are suggesting.

jwnasambu:
@hadijahkyampeire to be on the safe side let out wait for Ian's comments as you have suggested above since I believe he has a better understanding on the requirement.

src/apps/concepts/api.ts
ibacher: We definitely don't need a "generalFilters" that in the API (the other filters are actually sent directly to the backend; in this case we just need to change the `includeRetired` toggle which is already defined on line 40).

src/apps/concepts/pages/ViewConceptsPage.tsx
ibacher: I agree with @hadijahkyampeire on this point, but as JavaScript has a built-in boolean type, there's really no need to convert to numbers or anything.

```ts
includeRetired: initialGeneralFilters.includes('IncludeRetired'),
```

src/utils/constants.ts
```diff
@@ -305,3 +305,6 @@ export const CONTEXT = {
   view: "view",
   edit: "edit"
 };
+export const CONCEPT_GENERAL: string[] = [
```
ibacher, line 308: I'd almost think this was better suited for `src/app/components/constants.ts`.

hadijahkyampeire: Thanks, @ibacher just that I saw as if the backend expects number 1 or 0 🤔

ibacher: @hadijahkyampeire That would explain why changing `true` to `false` didn't do anything. I would think the best way to handle that would be to map from a boolean to an integer in `api.ts` (so when we make the call to the backend).

jwnasambu:
@ibacher and @hadijahkyampeire  thanks for the review and the comments. I have finished fixing the proposed changes  though am not seeing my work after `git push`

PR #675, opened 2021-03-19
OCLOMRS-954: Reset adding and deleting org members

hadijahkyampeire:
# JIRA TICKET NAME:
[Error Message Persists for failed Add Member](https://issues.openmrs.org/browse/OCLOMRS-954)

# Summary:
- Once the user clears the error message, it shouldn't show up again when I go back to that Organization, or when I go look at another different Organization or even the same organisation.

### DIFF
--- a/src/apps/organisations/pages/ViewOrgPage.tsx
+++ b/src/apps/organisations/pages/ViewOrgPage.tsx
@@ -13,7 +13,9 @@ import {
   addOrgMemberAction,
   addOrgMemberLoadingSelector,
   deleteOrgMemberAction,
-  deleteOrgMemberErrorSelector
+  deleteOrgMemberErrorSelector,
+  resetAddOrgMemberAction,
+  resetDeleteOrgMemberAction
 } from "../redux";
 import Header from "../../../components/Header";
 import { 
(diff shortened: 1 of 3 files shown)


src/apps/organisations/pages/ViewOrgPage.tsx
```diff
@@ -92,6 +98,11 @@ const ViewOrganisationPage: React.FC<Props> = ({
   
   const orgUrl = url.replace("/user", "").replace("edit/", "");
   
+  useEffect(() => {
+    resetAddOrgMember();
+    resetDeleteOrgMember();
+  }, []); // eslint-disable-line react-hooks/exhaustive-deps
```
BartlomiejRasztabiga, line 104: Shouldn't you add `resetAddOrgMember` and `resetDeleteOrgMember` methods to useEffect's hook dependency array as they can be changed in between renders?
Then you wouldn't have to disable `react-hooks/exhaustive-deps` rule

ibacher: Yeah, this isn't the nicest setup. The things that we really only want those functions to be executed on the initial render of the "page" component as they are responsible for clearing the Redux state of previous errors.

If you've got a way we can make that guarantee without disabling the eslint rule, I'd be open to it.

ibacher (approved):

PR #678, opened 2021-03-23
OCLMRS-919: added drag n drog functionality in AnswerMapping

nk183:
# JIRA TICKET NAME:
[JIRA ticket name](https://issues.openmrs.org/browse/OCLOMRS-919)

# Summary:
Added a Drag and Drop functionality in Answer Mapping 
added an extras property in the Mapping class which stores `{sort_wieght: number}` which will update when after re-ordering we click submit and be only available to `Custom Created Concept` on the `edit page` of concept

image


ibacher:
@nk183 One of the (very few) tests we have for this seems to be breaking. Would you be able to look into why that happens? You can run the tests locally (hopefully) by starting up the Docker image and running `npm run test:integration`.

PR #683, opened 2021-04-02
OCLOMRS-958:Errors from the backend should marked the field that caused the error

jwnasambu:
…ed the error

# JIRA TICKET NAME:
[Errors from the backend should marked the field that caused the error](https://issues.openmrs.org/browse/<OCLOMRS-958>)

# Summary: 
I modified the errorMsgReponse to returns an object rather than just a single string and added an useEffect() hook and update the form so that anything in the new error object field gets displayed at the bottom of the form.


HerbertYiga:
@jwamalwa  well done Julie, was this passing well locally?

jwnasambu:
@HerbertYiga yes locally it builds.

jwnasambu:
@ibacher Kind help please! Locally the changes build but when I make the PR the travis fails. Kindly how do I go about it?

src/redux/utils.ts
```diff
@@ -103,6 +103,10 @@ export function invalidateCache(action: string, dispatch: Function) {
 }
 
 export function errorMsgResponse(response: any) {
+  let errorMsgResponse = { __detail__: "Action could not be completed. Please retry." };
+
+  if (response.data && has(response.data, "detail")) {
+    errorMsgResponse["__detail__"] = response.data.detail
```
ibacher, line 109: The error in the build in Travis is caused by this line... essentially the `{` after the `if` is never closed. That said we don't need this `if` here at all. What you have on line 109 should be in place of what's on 117 and line 120-124 should look something like:

```ts
errorMsgResponse[key] =
    ? response.data[key].join(",")
    : response.data[key];
```

PR #689, opened 2021-04-15
OCLOMRS-949: Use concept url instead of version_url.

hadijahkyampeire:
# JIRA TICKET NAME:
[Concept links should not require a specific version](https://issues.openmrs.org/browse/OCLOMRS-949)

# Summary:
- Used concept URL instead of version_url


HerbertYiga:
@hadijahkyampeire  i find a some test is breaking ie apps/concepts/pages/tests/e2e/CreateOrEditConceptPage.test.ts

ibacher:
@hadijahkyampeire This seems a little bit different from what I was asking for in the ticket. We don't explicitly need to change any links. I just think we should render something for URLs like: https://openmrs.openconceptlab.org/orgs/CIEL/sources/CIEL/concepts/206/. Hopefully that just means adjusting a route somewhere.

hadijahkyampeire:
Oh I get it I will made the adjustments when we are ready to push this. Thanks @ibacher

hadijahkyampeire:
@ibacher I added another Route that can allow `conceptId` without version, I don't know if that solves the ticket, please review my PR.

ibacher (approved):
LGTM! Thanks @hadijahkyampeire!

PR #691, opened 2021-04-15
OCLOMRS-965:Set Source to whatever user want it to be (maybe “Your Recent Sources” instead of hardcoded)

jwnasambu:
# JIRA TICKET NAME:
[Set Source to whatever user want it to be (maybe “Your Recent Sources” instead of hardcoded)](https://issues.openmrs.org/browse/OCLOMRS-965)

# Summary:
I added `MySource` option on the public source  to enable a user add concepts  to the dictionary from his/her source.

### DIFF
--- a/src/utils/constants.ts
+++ b/src/utils/constants.ts
@@ -287,13 +287,22 @@ export const MAP_TYPE_CONCEPT_SET = option("CONCEPT-SET");
 
 const CIEL_SOURCE_URL = "/orgs/CIEL/sources/CIEL/";
 const PIH_SOURCE_URL = "/orgs/PIH/sources/PIH/";
+<<<<<<< HEAD
 const MSFOCP_SOURCE_URL = "/orgs/MSFOCP/sources/MSFOCP/"
+=======
+const MSFOCP_SOURCE_URL = "/orgs/MSFOCP/sources/MSFOCP/";
+const MySource_SOURCE_URL = "/orgs/MySource/sources/MySource/";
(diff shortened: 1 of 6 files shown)


src/utils/constants.ts
```diff
@@ -289,12 +289,14 @@ export const MAP_TYPE_CONCEPT_SET = option("CONCEPT-SET");
 const CIEL_SOURCE_URL = "/orgs/CIEL/sources/CIEL/";
 const PIH_SOURCE_URL = "/orgs/PIH/sources/PIH/";
 const MSFOCP_SOURCE_URL = "/orgs/MSFOCP/sources/MSFOCP/";
+const MySource_SOURCE_URL = "/orgs/MySource/sources/MySource/";
```
HerbertYiga, line 292: cc @hadijahkyampeire

jwnasambu: @HerbertYiga am missing your comment? I did that basing on this lick https://docs.google.com/presentation/d/1Q2yMyxTQrflQGc-O77KlGEXiPC_savon9ju9LDUaI5c/edit#slide=id.gd2498758a4_0_14 shared by Suruchi and the output is attached on the ticket.

hadijahkyampeire (changes requested):
@jwnasambu thanks for starting on these, but what is actually needed is not just the word `MySource` it was just an example that Suruchi use, it means to list all available user sources and public sources so that users can also pick from them rather than the hardcoded sources we currently have.

jwnasambu:
@hadijahkyampeire thanks for the clarification. Kindly, it sounds good when the tickets have clear description of what is expected if possible with diagram(s).

hadijahkyampeire:
You are right @jwnasambu I will ping the reporter to add more useful information that you can work with.

PR #695, opened 2021-04-23
OCLOMRS-964 : make header stick to the top

nk183:
# JIRA TICKET NAME:
[OCLOMRS-964](https://issues.openmrs.org/browse/OCLOMRS-964)

This functionality is not working for all the pages
this might be because of these:
- position sticky will most probably not work if overflow is set to hidden, scroll, or auto on any of the parents of the element.
- position sticky may not work correctly if any parent element has a set height.
but not able to found any of these above possibilities

https://user-images.githubusercontent.com/58779460/115831760-a7c42580-a42f-11eb-8020-fccad284e542.mp4

eg:
it is working on View Dictionary page
Not working o View Concept page

### DIFF
--- a/src/components/Header.tsx
+++ b/src/components/Header.tsx
@@ -23,7 +23,10 @@ import { Link, useHistory } from "react-router-dom";
 const useStyles = makeStyles((theme: Theme) =>
   createStyles({
     appBar: {
-      paddingLeft: theme.spacing(7) + 1
+      paddingLeft: theme.spacing(7) + 1,
+      position: "sticky",
+      top: "0",
+      width: "100%"
     },
     content: {
       marginTop: "6vh",


src/components/Header.tsx
```diff
@@ -23,7 +23,10 @@ import { Link, useHistory } from "react-router-dom";
 const useStyles = makeStyles((theme: Theme) =>
   createStyles({
     appBar: {
-      paddingLeft: theme.spacing(7) + 1
+      paddingLeft: theme.spacing(7) + 1,
+      position: "sticky",
```
HerbertYiga, line 27: @nk183 how does the UI appear on minimizing the page with these changes?

nk183: hi @HerbertYiga  it working the same 

https://user-images.githubusercontent.com/58779460/115845636-ac440a80-a43e-11eb-8dd4-eef3a0dc51ed.mp4

HerbertYiga: cc @ibacher

hadijahkyampeire:
@nk183 I have been trying out various options, I have something that kinda works but is hacky but maybe we can build on that. To the `Grid` of the `viewConceptsPage` the content class, if you add these styles it somehow works but they also need adjustments because position `fixed` messes up the margins.`bottom: 16px;
    height: 80%;
    overflow: auto;
    position: fixed;
    right: 71px;
    left: 90px;
    width: 95%;`

But there might be better options cc @ibacher

ibacher:
Here's my take on this: https://github.com/openmrs/openmrs-ocl-client/compare/OCLOMRS-964-IB. Basically, fix the weirdness on the `ViewConceptsPage.tsx` separately. I'm not sure if `marginTop: 32px` is a good idea or a bad idea.

nk183:
hi @ibacher,  https://github.com/openmrs/openmrs-ocl-client/compare/OCLOMRS-964-IB  is working great would be great if you proceed to merge it, closing this PR

PR #697, opened 2021-04-28
Improve Cypress

ibacher:
# Summary:

Overhauls the setup of Cypress and this application to be closer to Cypress's recommended setup (i.e., the e2e tests are in `cypress/integration` and we can directly interact with the Redux store).


hadijahkyampeire (approved):
This works well, thanks @ibacher, I noticed that you only left the `Login` tests and dropped the concepts and dictionary tests, was that intentional?

cypress.json
```diff
@@ -1,10 +1,7 @@
 {
   "baseUrl": "http://localhost:8080",
-  "integrationFolder": "./src",
-  "testFiles": ["**/tests/e2e/*.test.*", "**/tests/e2e/*.spec.*"],
-  "defaultCommandTimeout": 60000,
-  "requestTimeout": 60000,
-  "pageLoadTimeout": 60000,
+  "supportFile": "cypress/support/index.ts",
+  "testFiles": [ "LoginPage.spec.ts", "ViewConceptsPage.spec.ts" ],
```
hadijahkyampeire, line 4: Here you changed from generalizing to specifying filenames, won't that be too much, can't we just specify the directory, for example `**/cypress/integrations` since that's where all our e2e tests are moving to.

ibacher: Ah! Good catch! Earlier, I was just trying to leave the existing e2e tests in place and just focus on the one's I was converting. I actually think that we can just eliminate the `testFiles` property, since the default settings are "cypress/integration/**/*.*", which should work.

PR #700, opened 2021-05-06
OCLOMRS-958:Errors from the backend should marked the field that caused the error

jwnasambu:
# JIRA TICKET NAME:
[Errors from the backend should marked the field that caused the error](https://issues.openmrs.org/browse/OCLOMRS-958>)

# Summary:

### DIFF
--- a/src/apps/organisations/redux/actions.ts
+++ b/src/apps/organisations/redux/actions.ts
@@ -105,7 +105,7 @@ const retrieveOrganisationAction = (orgUrl: string) => {
 
         let errorMsg = errorMsgResponse(response);
 
-        const errorMessage =
+        const errorMessage: string | undefined =
           response?.data || response
             ? STATUS_CODES_TO_MESSAGES[response.status] || errorMsg
             : errorMsg?["__detail__"]:
--- a/src/apps/organisations/redux/actions.ts
+++ b/src/apps/organisations/redux/actions.ts
@@ -10,7 +10,6 @@ import {
 } from "../../../redux";
 import {
   getIndexedAction,
-  errorMsgResponse,
   FAILURE
 } from "../../../redux/utils";
 import {
@@ -30,7 +29,8 @@ import {
   HIDE_DELETE_MEMBER_DIALOG
 } from "./actionTypes";
 import { PERSONAL_ORGS_ACTION_INDEX } from "./constants";
-import { debug, STATUS_CODES_TO_MESSAGES } from "../../../utils";
+import { debug } from "../../../utils";
+import { AxiosResponse } from "axios";
 
 const createOrganisationAction = createActionThunk(
   CREATE_ORGANISATION_ACTION,
@@ -100,15 +100,8 @@ const retrieveOrganisationAction = (orgUrl: string) => {
       } catch (error) {
         debug(error, "redux/utils/#createActionThunk#:catch");
 
-        const response = error.response;
+        const response: AxiosResponse = error.response;
 
-
-        let errorMsg = errorMsgResponse(response);
-
-        const errorMessage: string | undefined =
-          response?.data || response
-            ? STATUS_CODES_TO_MESSAGES[response.status] || errorMsg
-            : errorMsg?["__detail__"]:
 
         dispatch({
           type: `${actionType}_${FAILURE}`,
@@ -186,7 +179,7 @@ const addOrgMemberAction = (
       } catch (error) {
         debug(error, "redux/utils/#createActionThunk#:catch");
 
-        const response = error.response;
+        const response: AxiosResponse = error.response;
 
         const errorMsgResponse: {[key: string]: string} = {
           "__detail__": "Action could not be completed. Please retry."
@@ -255,7 +248,7 @@ const deleteOrgMemberAction = (
       } catch (error) {
         debug(error, "redux/utils/#createActionThunk#:catch");
 
-        const response = error.response;
+        const response: AxiosResponse = error.response;
 
 
         const errorMsgResponse: {[key: string]: string} = {
--- a/src/apps/organisations/redux/actions.ts
+++ b/src/apps/organisations/redux/actions.ts
@@ -102,12 +102,13 @@ const retrieveOrganisationAction = (orgUrl: string) => {
 
         const response = error.response;
 
+
         let errorMsg = errorMsgResponse(response);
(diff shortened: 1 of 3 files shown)


src/redux/utils.ts
```diff
@@ -103,27 +103,22 @@ export function invalidateCache(action: string, dispatch: Function) {
...
+  for (let key in response.data) {
+    if (key === "__detail__") continue;
+    errorMsgResponse[key] =
         Array.isArray(response.data[key])
           ? response.data[key].join(",")
           : response.data[key]
-      );
     }
-  }
-  return errorMsgResponse.length > 0
```
ibacher, line 123: So, we need a `return errorMsgResponse;` line at the end of this.

src/apps/organisations/redux/actions.ts
```diff
@@ -102,12 +102,10 @@ const retrieveOrganisationAction = (orgUrl: string) => {
...
 
-        let errorMsg = errorMsgResponse(response);
 
-        const errorMessage: string | undefined | {} | [] =
-          response?.data || response
-            ? STATUS_CODES_TO_MESSAGES[response.status] || errorMsg
-            : errorMsg;
+        const errorMsgResponse: {[key: string]: string} = {
+          "__detail__": "Action could not be completed. Please retry."
+        }
```
ibacher, line 108: I forgot about these bits. This should be replaced with something like:

```ts
let errorMsg = errorMsgResponse(response);

const errorMessage =
  response?.data || response
    ? STATUS_CODES_TO_MESSAGES[response.status] || errorMsg
    : errorMsg["__detail__"];
```

ibacher (commented):
This is starting to shape up quite nicely I think. We still need to modify `getPrettyError()` [here](https://github.com/openmrs/openmrs-ocl-client/blob/315cadae009e97889056257574a37979e71c57ae/src/utils/utils.ts#L24) to replace `errors["__all__"]` with `errors["__detail__"]` and then test and ensure this works.

ibacher:
Already, so, to resolve the test errors we're getting, we need to actually change the tests in `src/redux/__test__/utils.test.ts` to reflect the new reality (this is fine and an expected part of this work).

Basically, what we need to do is change [this line](https://github.com/openmrs/openmrs-ocl-client/blob/a23b9bbcf8f0d0cc6b5948357dd2d7a1545ba6f8/src/redux/__test__/utils.test.ts#L30) from:

```ts
expect(errorMsgResponse(response)).toEqual(errorMsgs[error])
```

to:

```ts
expect(errorMsgResponse(response)["name"]).toEqual(errorMsgs[error])
```

src/redux/utils.ts
```diff
@@ -103,27 +103,23 @@ export function invalidateCache(action: string, dispatch: Function) {
...
-      );
-    }
-  }
-  return errorMsgResponse.length > 0
-    ? errorMsgResponse.join("\n")
-    : genericErrorMessage;
-}
+    } 
+    return errorMsgResponse;
+  };
```
ibacher, line 122:
```suggestion
  }

  return errorMsgResponse;
};
```

src/apps/organisations/redux/actions.ts
```diff
@@ -102,12 +102,13 @@ const retrieveOrganisationAction = (orgUrl: string) => {
 
         const response = error.response;
 
+
         let errorMsg = errorMsgResponse(response);
 
-        const errorMessage: string | undefined | {} | [] =
+        const errorMessage =
```
ibacher, line 108:
```suggestion
        const errorMessage: string | undefined =
```

ibacher (commented):
Hopefully those are the last little bits to get this building correctly.

src/apps/organisations/redux/actions.ts
ibacher: Alright... that was a silly suggestion.

Instead, let's change `errorMsgResponse()` to look like this:

```ts
export function errorMsgResponse(response: AxiosResponse) {
  const errorMsgResponse: {[key: string]: string} = {
    "__detail__": "Action could not be completed. Please retry."
  };

  if (response.data && Object.prototype.hasOwnProperty.call(response.data, "detail")) { 
    errorMsgResponse["__detail__"] = response.data.detail;
  } else if (response.status in STATUS_CODES_TO_MESSAGES) {
    errorMsgResponse["__detail__"] = STATUS_CODES_TO_MESSAGES[response.status];
  }
  
  for (let key in response.data) {
    if (key === "__detail__") continue;
    errorMsgResponse[key] =
          ? response.data[key].join(",")
          : response.data[key]
    } 
    
  return errorMsgResponse;
};
```

This also means hunting out the places where we call `errorMsgResponse()` and ensure that we cast the argument we're returning to an `AxiosResponse` (it should always be this), like so:

```ts
const response: AxiosResponse = error.response;
```

jwnasambu: Thanks @ibacher let me do it right a way

ibacher: @jwnasambu So the second part of this is to just eliminate anything that looks like this:

```
const errorMessage =
  response?.data || response
    ? STATUS_CODES_TO_MESSAGES[response.status] || errorMsg
    : errorMsg["__detail__"];
```

From the code base... actually anything that processes the `errMsg` after it's returned from `errorMsgResponse(Response)`. The idea here is that that processing will be done on the actual page itself.

jwnasambu: @ibacher thanks for the clear explanation though it seems I had eliminated that part already!

jwnasambu: @ibacher am sorry I have just seen that part and other proposed changes. Kindly let me fix and push the changes right away.

jwnasambu:
@hadijahkyampeire Kindly open this PR we were still making some changes on it with Ian.

hadijahkyampeire:
@jwnasambu I am sorry but I am unable to reopen it, its repository was deleted you will need to open a new PR.

jwnasambu:
@ibacher should I open a new PR or the PR can be re-opened again? Thanks for your help.

ibacher:
@jwnasambu If you have a local copy of your OCLOMRS-958 branch, you should be able to push it to GitHub. Right now, that branch doesn't exist on your GitHub repo, which is why this can't be reopened.

PR #701, opened 2021-05-10
OCLOMRS-965:Set Source to whatever user want it to be (maybe “Your Recent Sources” instead of hardcoded)

jwnasambu:
# JIRA TICKET NAME:
[Set Source to whatever user want it to be (maybe “Your Recent Sources” instead of hardcoded)](https://issues.openmrs.org/browse/OCLOMRS-965)

# Summary:

### DIFF
--- a/src/apps/concepts/components/ViewConceptsHeader.tsx
+++ b/src/apps/concepts/components/ViewConceptsHeader.tsx
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { useState } from "react";
 import { Link } from "react-router-dom";
 import Header from "../../../components/Header";
 import {
@@ -13,6 +13,7 @@ import {
   makeStyles,
   Menu,
   MenuItem,
+  TextField,
   Theme
 } from "@material-ui/core";
 import { PREFERRED_SOURCES_VIEW_ONLY, useAnchor } from "../../../utils";
@@ -31,6 +32,23 @@ const useStyles = makeStyles((theme: Theme) =>
   createStyles({
     lightColour: {
       color: theme.palette.background.default
+    },
+    textField: {
+      padding: "0.2rem 1rem",
+      cursor: "none"
+    },
+    input: {
+      cursor: "pointer",
+      borderBottom: "1px dotted black",
+      paddingBottom: "0.25rem"
+    },
+    underline: {
+      "&&&:before": {
+        borderBottom: "none"
+      },
+      "&&:after": {
+        borderBottom: "none"
+      }
     }
   })
 );
@@ -43,6 +61,8 @@ const ViewConceptsHeader: React.FC<Props> = ({
   children,
   sources
 }) => {
+  const [showSources, setShowSources] = useState(false);
+  const formatPrefferedSources = Object.entries(PREFERRED_SOURCES_VIEW_ONLY).map(([key, value]) => ({ name: key , url: value }));
   const classes = useStyles();
   const isSourceContainer = containerType === SOURCE_CONTAINER;
   const isAddToDictionary = isSourceContainer && !!addConceptToDictionary;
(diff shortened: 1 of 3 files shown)


hadijahkyampeire (commented):
This looks good @jwnasambu thanks, the only thing left is adding preferred sources to the list I think, I will add a suggestion on this PR and also confirm with @gracepotma about it.

src/apps/concepts/components/ViewConceptsHeader.tsx
```diff
@@ -38,7 +40,8 @@ const ViewConceptsHeader: React.FC<Props> = ({
   containerUrl,
   gimmeAUrl,
   addConceptToDictionary,
-  children
+  children,
+  sources
 }) => {
   const classes = useStyles();
```
hadijahkyampeire, line 46: We need to first format the preferred sources list so that it looks almost like our other sources list.
`const formattedPrefferedSources = Object.entries(PREFERRED_SOURCES_VIEW_ONLY).map(([key, value]) => ({ name: key , url: value }));`

src/apps/concepts/components/ViewConceptsHeader.tsx
```diff
@@ -77,18 +80,18 @@ const ViewConceptsHeader: React.FC<Props> = ({
           open={Boolean(switchSourceAnchor)}
           onClose={handleSwitchSourceClose}
         >
-          {Object.entries(PREFERRED_SOURCES_VIEW_ONLY).map(
-            ([preferredSourceName, preferredSourceUrl]) => (
+           {sources?.map(
```
hadijahkyampeire, line 83:
```suggestion
           {formattedPrefferedSources.concat(sources)?.map(
```

hadijahkyampeire (commented):
Let me know if you have any questions about this.

src/apps/concepts/pages/ViewConceptsPage.tsx
```diff
@@ -381,7 +396,8 @@ const mapActionsToProps = {
   retrieveDictionary: makeRetrieveDictionaryAction(true),
   retrieveSource: retrieveSourceAndDetailsAction,
   addConceptsToDictionary: recursivelyAddConceptsToDictionaryAction,
-  removeConceptsFromDictionary: removeReferencesFromDictionaryAction
+  removeConceptsFromDictionary: removeReferencesFromDictionaryAction,
+  retrieveSources: retrievePublicSourcesAction
```
ibacher, line 400: I would reuse the `retrievePublicSources` API that already exists in `src/apps/concepts/api.ts` for this. Despite it name, it actually will get all of the sources the current user has access to. This is the same thing we use to get the sources in the `MappingsTableRow.tsx` file.

It actually might be a good idea to wrap this up in it's own Redux action and Thunk so that we can cache the results somewhere instead of needing to load this list everytime, but that doesn't need to be part of this ticket.

I haven't tried this out, but it looks like it does the right things.

jwnasambu: > It actually might be a good idea to wrap this up in it's own Redux action and Thunk so that we can cache the results somewhere instead of needing to load this list everytime, but that doesn't need to be part of this ticket.
> 
> I haven't tried this out, but it looks like it does the right things.

Is it okay to put it the the follow up PR of this ticket https://issues.openmrs.org/browse/OCLOMRS-967?

hadijahkyampeire:
Hey @jwnasambu , how far with the feedback on this work? let me know if you need any help so that we can merge the work.

ibacher:
I'm actually happy with this being merged and my feedback becoming a new ticket. Something is better than nothing.

hadijahkyampeire:
Great, I will approve and merge it then.

hadijahkyampeire (approved):
Great work!

jwnasambu:
> I'm actually happy with this being merged and my feedback becoming a new ticket. Something is better than nothing.

@ibacher I will be the happiest person working on the feedback ticket that you will create. Am fixing [OCLOMRS-967](https://issues.openmrs.org/browse/OCLOMRS-967) which moves hand in hand with this ticket.

PR #708, opened 2021-06-15
OCLOMRS-998: Add Trusted content flags MVP

suruchee:
# Add Trusted content flags MVP:
[Add Trusted content flags MVP](https://issues.openmrs.org/browse/OCLOMRS-998)

# Summary:
MVP for the Trusted content flag so added a constant CIEL and marked it as a trusted one by adding Verified User Icon.


src/apps/containers/components/ContainerSearch.tsx
```diff
@@ -32,45 +36,73 @@ const useStyles = makeStyles((theme: Theme) =>
...
+    flex: {
+      display: "flex",
+    },
   })
 );
 
 const ContainerSearch: React.FC<Props> = ({ title, onSearch, initialQ }) => {
   const classes = useStyles();
   const [q, setQ] = useState(initialQ);
+  const [state, setState] = React.useState({
```
hadijahkyampeire, line 50:
```suggestion
  const [verified, displayVerified] = useState(false);
```

src/apps/containers/components/ContainerSearch.tsx
```diff
@@ -32,45 +36,73 @@ const useStyles = makeStyles((theme: Theme) =>
...
+    displayVerified: false,
+  });
+  const handleChange = (event: any) => {
+    setState({ ...state, [event.target.name]: event.target.checked });
```
hadijahkyampeire, line 54:
```suggestion
    displayVerified(event.target.checked);
```

src/apps/containers/components/ContainerSearch.tsx
```diff
@@ -32,45 +36,73 @@ const useStyles = makeStyles((theme: Theme) =>
...
+                </IconButton>
+              </InputAdornment>
+            }
+          />
+        </form>
+        <FormControlLabel
+          control={
+            <Switch
+              checkedIcon={<VerifiedUser />}
+              checked={state.displayVerified}
```
hadijahkyampeire, line 93:
```suggestion
              checked={verified}
```

src/apps/containers/components/ContainerSearch.tsx
```diff
@@ -32,45 +36,73 @@ const useStyles = makeStyles((theme: Theme) =>
...
+              onChange={handleChange}
+              color="primary"
+              name="displayVerified"
+            />
+          }
+          label={
+            state.displayVerified
```
hadijahkyampeire, line 100:
```suggestion
            verified
```

hadijahkyampeire (approved):
Great work @suruchee , UI looks great, code is great too, just some minor suggestions to keep the state simple, dealing with objects sometimes can be tricky so since that is a true/false state you can just keep it simple.

hadijahkyampeire:
@ibacher take a look at this when you get some time.

src/apps/containers/components/ContainerCard.tsx
```diff
@@ -47,53 +60,62 @@ const ContainerCard: React.FC<Props> = ({
...
+              color="primary"
+              data-testid={`viewButton-${index}`}
+            >
+              View
+            </Button>
+          </CardActions>
+        </div>
+        <div>
+          {name === "CIEL" ? (
+            <VerifiedUser className={classes.trustedIcon} />
```
ibacher, line 114: Instead of defining this in two places with separate classes, this should be turned into a component in `src/components`, e.g., a `VerifiedSource` component.

src/apps/containers/components/ContainerCard.tsx
```diff
@@ -47,53 +60,62 @@ const ContainerCard: React.FC<Props> = ({
...
+              variant="text"
+              color="primary"
+              data-testid={`viewButton-${index}`}
+            >
+              View
+            </Button>
+          </CardActions>
+        </div>
+        <div>
+          {name === "CIEL" ? (
```
ibacher, line 113: Shouldn't this be checking against the `VERIFIED_SOURCES` constant?

src/apps/containers/components/ContainerSearch.tsx
```diff
@@ -32,45 +36,73 @@ const useStyles = makeStyles((theme: Theme) =>
...
+                >
+                  <SearchIcon />
+                </IconButton>
+              </InputAdornment>
+            }
+          />
+        </form>
+        <FormControlLabel
+          control={
+            <Switch
```
ibacher, line 91: Several things:

1. Does it make sense to have this toggle appear on *every* source screen? E.g., is there a use case for limiting things to "My verified sources" or should this just be limited to the public sources / dictionaries screen?
2. The label is relatively long and seems quite obtrusive and not fitting with Material design standards. Could we convert the label into, e.g., a tool tip or something less obtrusive? (Notice how Switches in [the Material docs](https://material.io/components/switches) are laid-out vertically rather than squashed on a line with other components as is done here. Maybe moving this to the line above the search box?
3. Should the state of this toggle persist as I change through views (i.e., when I go to "My Sources" and switch it on and then go to "Public Sources" shouldn't it still remain on?)
4. See screenshot below:
  
Screenshot showing toggle with Public Sources visible

It seems weird to me that toggling this on is showing me what appear to be non-verified sources (I think those SNOMED sources are part of CIEL, but they aren't actually verified).

ibacher (changes requested):
Thanks for this great work Suruchi!

In addition to the below comments, one thing, if you're going to use Prettier to format the sources, please use the version included in the Dictionary Manager package.json. Otherwise, these PRs get a bit noisy!

suruchee:
Thank you @ibacher and @hadijahkyampeire  for the review and comment. I will update with all the mentioned comments.

src/apps/containers/components/ContainerSearch.tsx
suruchee: @ibacher, I actually made changes with respect to design. @gracepotma, will you please add comments, how we can move forward?

suruchee: Thank you @ibacher  for the comment, I will update once I receive comments from Grace.

gracepotma: 1. I see what you're saying Ian but personally I've been finding that I create a number of fake dictionaries. It would be nice to easily filter these out, e.g. I notice the same thing happening for our main OHRI user when he's demoing and sharing his screen, his real work/real dictionaries are cluttered by practice/in-progress/not-prime-time-titled dictionaries. But, if it's easier to just apply this toggle to the Public tab, then yes, that's fine.

gracepotma: 2. @ibacher I agree this isn't quite the optimal design - if we wanted to design for possibly having more filter options in the future, we'd have an additional step/UI workflow for filtering - but for the sake of getting this out into the world more conceptually I am okay with this hack for now. (I know, slippery slope, so I do appreciate you raising this, as we'll need to protect the app from too many hacky UX practices.) The alternative that we played with in the sketches looked like this, do you prefer this layout? https://docs.google.com/presentation/d/1Q2yMyxTQrflQGc-O77KlGEXiPC_savon9ju9LDUaI5c/edit#slide=id.gda809a9155_1_38

gracepotma: 3. Yes, I agree it would be nice for the toggle state to persist. But this is a nice-to-have and could be added later, unless @suruchee this is easy to add to this PR.

gracepotma: 4. Agreed. @suruchee to work on this (turns out it's due to a hard coded filter to find things labeled with "CIEL" which those happen to have)

ibacher: @grace Re: 2. I personally prefer the layout with the toggle underneath the search, but I agree that it shouldn't be a blocker for this PR.

gracepotma: Thanks @ibacher, yeah I actually prefer that layout too. @suruchee if that's easy to change go ahead, but if not please proceed :)

ibacher: @gracepotma Re: 1. I totally agree that a toggle to show my "real" dictionaries is very useful. I think I mostly just asked that since, for now, we only have a hard-coded list of sources that we're checking against so toggling the filter will *always* filter out all the sources on the My Sources page. Again, not a blocker, just seems like an... unexpected UX.

ibacher:
@suruchee Do you need anything else to make the requested changes? Otherwise, I'm excited to have this feature added.

src/apps/containers/components/ContainerSearch.tsx
gracepotma: Oh yeah good point @ibacher, you're totally right 🤦  Well, let's leave it for now, so that we don't have to worry about adding it back in later.

suruchee:
@ibacher, thanks for the comment. I have pushed with the update but, for comment no. 3 and 4 I could not work on toggle persist and showing of non verified one as I have used q=CIEL as queryParams and it gave 3 results. And for 3.  every time tab is switched, it gets reloaded so can you please help me in this regard?

ibacher:
@suruchee So, re: 3, to make the state persistent across page views, we need the state to be persistent across renders (since every click on the tab triggers a new React route and thus a new render).

Given the current state of the application, the easiest way to do this is to store the toggle state in the Redux store rather than in a component-local `useState()` hook.

Currently, we don't have a Redux reducer for the containers app, but rather we separately track state for dictionaries and sources. So it's probably fine to just add a toggle to those states.

To do this, you'll need to add a new property to the initial state (which is setup in `/apps/<dictionaries or sources>/redux/reducer.ts`, e.g. [here for dictionaries](https://github.com/openmrs/openmrs-ocl-client/blob/b132ecc97a62e7cf56c2ce73302de24cc30cdbde/src/apps/dictionaries/redux/reducer.ts#L18). This is as easy as adding something like:

```ts
const initialState: DictionaryState = {
  dictionaries: [],
  versions: [],
  addReferencesResults: [],
  showOnlyVerified: false
};
```

Then we need to add some ceremony to toggle this state. The ceremony involved means:

1. Creating an actionType (in `/redux/actionTypes.ts`) for the new action, e.g.
    ```ts
     export const TOGGLE_SHOW_VERIFIED_ACTION = "dictionary/toggleShowVerified";
    ```
2. Adding the action to the reducer to actually do the toggling [that's this construct](https://github.com/openmrs/openmrs-ocl-client/blob/b132ecc97a62e7cf56c2ce73302de24cc30cdbde/src/apps/dictionaries/redux/reducer.ts#L23), which might look something like:
    ```ts
    [TOGGLE_SHOW_VERIFIED_ACTION]: (state, action) => ({
      ...state,
      showOnlyVerified: !state.showOnlyVerified
    })
    ```
3. Adding an actual action to trigger the reducer (this goes in `/redux/actions.ts`):
    ```ts
    export const toggleShowVerifiedAction = () => {
      return (dispatch: Function) => dispatch(TOGGLE_SHOW_VERIFIED_ACTION);
    }
4. Wiring the reducer into the page components. This means going into each of the `View...DictionariesPage.tsx` files and modifying `mapStateToProps` to something like this:
    ```ts
    const mapStateToProps = (state: AppState) => ({
      loading: retrieveOrgDictionariesLoadingSelector(state),
      dictionaries:
        state.dictionaries.dictionaries[ORG_DICTIONARIES_ACTION_INDEX]?.items,
      meta:
        state.dictionaries.dictionaries[ORG_DICTIONARIES_ACTION_INDEX]?.responseMeta,
      showOnlyVerified: state.dictionaries.showOnlyVerified
    });
     ```
    And `mapDispatchToProps` to something like this:
    ```ts
    const mapDispatchToProps = {
      retrieveDictionaries: retrieveOrgDictionariesAction,
      toggleShowVerified: toggleShowVerifiedAction
    };
    ```
5. Then modifying `ViewDictionariesPage` and `ViewDictionaries` to take these as arguments and pass them down (`ViewDictionariesPage` should pass them to `ViewDictionaries`, `ViewDictionaries` should pass them as properties to `ContainerSearch`)
6. Modify `ContainerSearch` to use the new `showOnlyVerified` and `toggleShowVerified` things instead of `useState()` and `handleChange()`. This basically means that `showOnlyVerified` replaces `verified` [in this line](https://github.com/openmrs/openmrs-ocl-client/pull/708/files#diff-d844cd27ed11475b4447913223fb3b1a2ed1d7b0075aa93658e6ffcdfe3b8f6dR88) and `toggleShowVerified` replaces `handleChange` [in this one](https://github.com/openmrs/openmrs-ocl-client/pull/708/files#diff-d844cd27ed11475b4447913223fb3b1a2ed1d7b0075aa93658e6ffcdfe3b8f6dR89)
7. Add a `useEffect()` hook to `ContainerSearch` to request the right objects. Something like this should work:
    ```ts
    useEffect(() =>
      showOnlyVerified ? onSearch(VERIFIED_SOURCES[0]) : onSearch(),
      [showOnlyVerified]
    );
    ```

Sorry for the mass of instructions. Hopefully that's clear enough.

ibacher:
Re 4: I actually don't think that kind of filtering is possible \*sigh*.

suruchee:
Thank you so much @ibacher, for the clear description and instruction. I will work on it and update soon.

suruchee:
Thank you @ibacher  for your support and guidance. I updated removing the warnings and now everything looks fine.

ibacher (approved):

PR #709, opened 2021-06-16
OCLOMRS-989:Create a feature file ready for automated test

jwnasambu:
# JIRA TICKET NAME:
[Create a feature file ready for automated test](https://issues.openmrs.org/browse/OCLOMRS-989)

# Summary:
I created feature files ready for automated testing.

### DIFF
--- a/cypress/integration/dictionary.feature
+++ b/cypress/integration/dictionary.feature
@@ -1,0 +1,15 @@
+Feature: Dictionary Manager
+  Background:
+
+
+  Scenario: A user should be able to Create new dictionary
+  Given the user is logged in
+   Given the user clicks on Dictionaries icon 
+   Given the user clicks on create new dictionary icon
+   Then the system loads the Create dictionary form
+    When the user clicks on Create new dictionary icon
+    When the user enters the required information in the given fields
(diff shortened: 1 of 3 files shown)


hadijahkyampeire:
@jwnasambu, thanks for breaking the ground ❤️ ,  I think you should modify this structure to align with the other automated tests for login that we already have i.e instead of first adding folders and the files, just add files to be consistent with what we already have.

Also if it's not too much work, add some scenarios to one or two of the feature files rather than leaving all of them empty, we can modify them if we need to in the near future.

hadijahkyampeire:
Refer to this PR for the structuring https://github.com/openmrs/openmrs-ocl-client/pull/705

jwnasambu:
@hadijahkyampeire thanks for the review and I really get what you are pointing out though I was just following what the ticket requires which I may be wrong on the way I understood it.

jwnasambu:
> Also if it's not too much work, add some scenarios to one or two of the feature files rather than leaving all of them empty, we can modify them if we need to in the near future.

This has its in depended ticket(s) though am already working on one of the tickets though I haven't claimed the ticket.

PR #710, opened 2021-06-17
OCLOMRS-933: Add ability to include and exclude already added concepts.

hadijahkyampeire:
# JIRA TICKET NAME:
[General Filter: Hide Added Concepts & Add Filter option to Include Added Concepts
](https://issues.openmrs.org/browse/OCLOMRS-933)

# Summary:
- Added ability to include and exclude.
## Caveats
- Used the endpoint provided by backend that returns only concepts that are not yet added, refer to this PR https://github.com/openconceptlab/ocl_issues/issues/728

### DIFF
--- a/src/apps/concepts/api.ts
+++ b/src/apps/concepts/api.ts
@@ -32,6 +32,7 @@ const api = {
       classFilters?: string[];
       sourceFilters?: string[];
       includeRetired?: boolean;
+      includeAdded?: boolean;
     }) => {
       const {
         conceptsUrl = "#",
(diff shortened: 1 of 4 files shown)


hadijahkyampeire:
Some screenshots showing when one excludes or includes already added concepts. By default, the page returns only concepts that

 are not already added.
Screenshot 2021-06-17 at 15 54 32

- When include Added is true
Screenshot 2021-06-17 at 15 54 56

ibacher (approved):

PR #711, opened 2021-06-24
OCLOMRS-989: Create a feature file for Create dictionary for automated test

jwnasambu:
# JIRA TICKET NAME:
[Create a feature file for Create dictionary for automated test](https://issues.openmrs.org/browse/OCLOMRS-989)

# Summary:
 I created feature files  and step description for create dictionary ready for automated testing.


cypress/integration/Dictionary/createDictionary.feature
```diff
@@ -0,0 +1,110 @@
+Feature: Dictionary Management
+
+  Scenario: A user to create a new dictionary
```
ibacher, line 3:
```suggestion
  Scenario: A user can create a new dictionary
```

I'd similarly replace "should be able to" with "can"

cypress/integration/Dictionary/createDictionary.feature
```diff
@@ -0,0 +1,110 @@
...
+    Given the user is logged in
```
ibacher, line 4: Since all of these tests assume the user is logged in, this should be something that is setup in the background [as is demonstrated in the "Login Page"](https://github.com/openmrs/openmrs-ocl-client/blob/4d5150a2a475cba0e7876bf77c1aa39c54aaee75/cypress/integration/login.feature#L2) feature. I.e., you should have something like:

```gherkin
Background:
  Given the user is not logged in
```

cypress/integration/Dictionary/createDictionary.feature
```diff
@@ -0,0 +1,110 @@
...
+    And is on the user collections page
+    Then the user clicks on Create Dictionary Icon
+     Then the user should be able to create new dictionary
```
ibacher, line 7: NB I know this style is a little difficult to write in because the commands are actually aligned by their last letter rather than by leading space. The idea, though, is that this way, the relevant text (i.e. the parts beginning "the user") all start at the same column, which makes it a bit visually easier to see.

```suggestion
    Given the user is on the user collections page
     Then the user clicks on Create Dictionary icon
     Then the user should sent to the Create Dictionary page
```

cypress/integration/Dictionary/createDictionary.feature
```diff
@@ -0,0 +1,110 @@
...
+
+  Scenario: A user should be able to create new dictionary by pressing on Create new dictionary icon
+    Given the user is on the Dictionaries page
+     When the user clicks on Create new dictionary icon
+     Then the create Dictionary form is filled
+     Then the user should be able to create new dictionary
```
ibacher, line 13: This doesn't seem like a useful scenario to have. The first part, i.e., that clicking the icon sends the user to the page is tested in the first scenario. The second part, i.e., that the dictionary can be created is tested in other scenarios.

cypress/integration/Dictionary/createDictionary.feature
```diff
@@ -0,0 +1,110 @@
...
+  
+  Scenario: A user should be able to create a new dictionary by filling the required information
+    Given the user is on the Create Dictionary page
+     When the user enters the required information in the form
+      And the user clicks the submit button
+     Then the user should be able to create a new dictionary
+  
+  Scenario: The Create Dictionary form should require a Dictionary Name to Create new dictionary
```
ibacher, line 21: It's best to use consistent words to describe scenarios. Since everything else is "A user can...", this should be something like "A user must supply a dictionary name".

cypress/integration/Dictionary/createDictionary.feature
```diff
@@ -0,0 +1,110 @@
...
+    Given the user is on the Create Dictionary page
+     When the user enters their Dictionary Name
```
ibacher, line 23: Don't needlessly capitalises things

```suggestion
     When the user enters their dictionary name
```

cypress/support/step_defination/createDictionary/createDictionary.ts
```diff
@@ -0,0 +1,153 @@
+/// <reference types="cypress" />
+/// <reference types="../../" />
+import { Given, When, Then } from "cypress-cucumber-preprocessor/steps";
+
+Given('the user is logged in', () => {
+  cy.login();
+});
+
+Given(' the user is on the Dictionaries page', () => {
```
ibacher, line 9: Leading spaces are significant, so watch out for those

```suggestion
Given('the user is on the Dictionaries page', () => {
```

cypress/integration/Dictionary/createDictionary.feature
```diff
@@ -0,0 +1,110 @@
...
+      And the user submits the form
+     Then the user should still be on the Create Dictionary page
+      And the Short code, Preferred Source, Owner, Visibility and Preferred Language fields should be marked as having an error
+      And the error message "Short code is required" should be visible
+      And the error message "Select a preferred source" should be visible
+      And the error message "Select this dictionary's owner" should be visible
+      And the error message "Select who will have access to this dictionary" should be visible
+      And the error message "Select a preferred language" should be visible
```
ibacher, line 31: This seems like a backwards way to test what we want to test here. What the scenario says is that we require a name to create a dictionary, so the way to test this is to fill in the form with everything *except* the name and then ensure that the *name* field is marked as an error. In other words, we want something like:

```gherkin
  Given the user is on the Create Dictionary page
   When the user enters the required values except name
    And the user submits the form
   Then the user should still be on the Create Dictionary page
    And the name field should be marked as having an error
    And the error message "Name is required" should be visible
```

cypress/integration/Dictionary/createDictionary.feature
```diff
@@ -0,0 +1,110 @@
...
+  
+  Scenario: The Create Dictionary form should require a Short Code to Create new dictionary
```
ibacher, line 33: So a word on capitalization: I think it's fine to describe pages and forms with capitals because when you go the the "create dictionary" page, it says "Create Dictionary" on the top. Otherwise, capitalisation should generally be avoided except where, e.g., at the start of a scenario (which is really a sentence).

```suggestion
  Scenario: The Create Dictionary form should require a short sode to create new dictionary
```

ibacher (changes requested):
@jwnasambu Some good stuff here! I've left a number of comments. However, overall, instead of trying to write the full feature file and then write all of the code to test the feature file, I would do this iteratively. What that means is:

1. Write a scenario (e.g. "A user can create a new dictionary by clicking on the create dictionary button")
2. Write the steps for the scenario
3. Write the implementation for the scenario
4. Ensure the scenario works locally
5. Move on to the next scenario

Despite how things wound up looking, this is how I wrote the login feature. So the first version looked like this:

```gherkin
Feature: Login Page

  Scenario: A user should be able to login
    Given the user is not logged in
      And the user is on the login page
     When the user enters their credentials
      And the user submits the form
     Then the user should be logged in
```

With a corresponding implementation:

```ts
/// <reference types="cypress" />
/// <reference types="../../" />
import { Given, When, Then } from "cypress-cucumber-preprocessor/steps";

// ensure the user is logged out
Given("the user is not logged in", () => {
  cy.logout();
});

Given("the user is on the login page", () => {
  cy.visit("/login");
});

When("the user enters their credentials", () => {
  cy.get("#username").type("admin");
  cy.get("#password").type("Admin123");
});

When("the user submits the form", () => {
  cy.get("#login-form form").submit();
});

Then("the user should be logged in", () => {
  cy.url().should("not.contain", "/login");
});
```

Then I wrote the second scenario:

```gherkin
  Scenario: A user should be able to login by pressing enter
    Given the user is not logged in
      And the user is on the login page
     When the user enters their credentials
      And the user presses enter
     Then the user should be logged in
```

But I realised I was probably going to be having the `Given the user is not logged in` line everywhere at that point which felt quite repetitive, so I [consulted the Gherkin reference](https://cucumber.io/docs/gherkin/reference/) and discovered [there was a feature to handle repetitive `Given` commands](https://cucumber.io/docs/gherkin/reference/#background), so my feature file now looked like:

```gherkin
Feature: Login Page
  Background:
    Given the user is not logged in
      And the user is on the login page

  Scenario: A user should be able to login
     When the user enters their credentials
      And the user submits the form
     Then the user should be logged in

  Scenario: A user should be able to login by pressing enter
     When the user enters their credentials
      And the user presses enter
     Then the user should be logged in
```

(Note that I later removed the `And` part from the Background when I realised not every single one of my tests needed to start on the login page).

Now, even though I've added a whole bunch of lines to the scenario, there's really only one thing that needs to be implemented, namely the line `And the user presses enter` (all of the other steps are already implemented for the first test). So, in my implementation, I added:

```ts
When("the user presses enter", () => {
  cy.type("{enter}");
});
```

This didn't work... Cypress gave me some sort of error about needing a "target", so I changed it to:

```ts
When("the user presses enter", () => {
  cy.get("#password").type("{enter}");
});
````

Figuring that when the user presses "enter" in the appropriate way it's most likely after they've just typed their password.

Basically, instead of trying to write everything in one go, just write one scenario, get it to work, move on to the next scenario, implement as little as you can to get it to work, etc. And continuing like this, you can wind up with something that looks quite complex, but was really just done step-by-step.

@hadijahkyampeire @suruchee (and any others) hopefully this description of the process I went through to develop these tests is helpful.

cypress/integration/Dictionary/createDictionary.feature
jwnasambu: @ibacher thanks for the review! Let me fix the proposed changes.

suruchee:
As the expected test for this issue is covered by https://github.com/openmrs/openmrs-ocl-client/pull/713 , can you please close this PR.

jwnasambu:
@suruchi am not understanding your comment please!

jwnasambu:
@suruch I have seen the changes Ian incorporated in what I did.Thanks for the point out.

hadijahkyampeire:
@ibacher already addressed this.

PR #712, opened 2021-06-30
OCLOMRS-967: Pick concepts from a dictionary (in addition to sources)

jwnasambu:
# JIRA TICKET NAME:
[Pick concepts from a dictionary (in addition to sources)](https://issues.openmrs.org/browse/OCLOMRS-967)

# Summary:
I worked on a feature that allows a user to create a dictionary that contains concepts from another dictionary.


src/apps/concepts/components/ViewConceptsHeader.tsx
```diff
@@ -114,49 +123,32 @@ const ViewConceptsHeader: React.FC<Props> = ({
           <TextField 
             multiline
             className={classes.textField} 
-            InputProps={{
-              className: classes.underline
-            }}
```
ibacher, line 119: Why was this removed?

src/apps/concepts/components/ViewConceptsHeader.tsx
```diff
@@ -36,24 +38,21 @@ const useStyles = makeStyles((theme: Theme) =>
...
       paddingBottom: "0.25rem"
     },
-    underline: {
-      "&&&:before": {
-        borderBottom: "none"
-      },
-      "&&:after": {
-        borderBottom: "none"
-      }
-    },
```
ibacher, line 53: I assume this should be restored as well?

src/apps/concepts/components/ViewConceptsHeader.tsx
```diff
@@ -36,24 +38,21 @@ const useStyles = makeStyles((theme: Theme) =>
...
     sourceIcon:{
       marginRight: "0.2rem",
       fill: "#8080809c"
+    },
+    dictionaryIcon:{
+      marginRight: "0.2rem",
+      fill: "#8080809c"
```
ibacher, line 55: We don't really need a separate `dictionaryIcon` class... This can be shared with `sourceIcon` (at least until we develop a need to style the icons differently)

src/apps/concepts/components/ViewConceptsHeader.tsx
```diff
@@ -36,24 +38,21 @@ const useStyles = makeStyles((theme: Theme) =>
     },
     textField: {
       padding: "0.2rem 1rem",
-      cursor: "none"
+      cursor: "none",
+      width: "100%"
```
ibacher, line 42: Why this change?

src/apps/dictionaries/redux/actions.ts
```diff
@@ -391,11 +392,19 @@ export const recursivelyAddConceptsToDictionaryAction = (
     );
 
     const referencesToAdd = await recursivelyFetchToConcepts(
-      sourceUrl,
       concepts.map(concept => concept.id),
-      updateProgress
+      updateProgress,
+      sourceUrl,
+      false
```
ibacher, line 398: This is in the wrong order. `sourceUrl` should be the last argument.

src/apps/dictionaries/redux/actions.ts
```diff
@@ -354,10 +355,10 @@ export const editDictionaryVersionAction = createActionThunk(
 );
 
 export const recursivelyAddConceptsToDictionaryAction = (
-  sourceUrl: string,
   dictionaryUrl: string,
   rawConcepts: (APIConcept | string)[],
-  bulk: boolean = false
+  bulk: boolean = false,
+  sourceUrl?: string
```
ibacher, line 361: What happens if `sourceUrl` is blank but one or more `rawConcepts` is a string? We should raise an error in that case.

src/apps/dictionaries/redux/actions.ts
```diff
@@ -391,11 +392,19 @@ export const recursivelyAddConceptsToDictionaryAction = (
...
     const referencesToAdd = await recursivelyFetchToConcepts(
-      sourceUrl,
       concepts.map(concept => concept.id),
-      updateProgress
+      updateProgress,
+      sourceUrl,
+      false
     );
 
+    const groupedConcepts = groupBy(concepts.sourceUrl),
```
ibacher, line 401:
```suggestion
    const groupedConcepts = groupBy(concepts.sourceUrl);
```

src/apps/dictionaries/redux/actions.ts
ibacher: Also, we should do away with *this* copy of `recursivelyFetchToConcepts()` since we're using that below, right?

src/apps/dictionaries/redux/actions.ts
```diff
@@ -391,11 +392,19 @@ export const recursivelyAddConceptsToDictionaryAction = (
...
+     const referencesToAdd = Object.entries(groupedConcepts).map(([source, concepts]) => recursivelyFetchToConcepts(
+       concepts.sourceUrl,
```
ibacher, line 403: The `sourceUrl` should be the `source` argument. Remember, however, that the `sourceUrl` now needs to be the *last* argument to this, if we're re-arranging the argument order for `recursivelyFetchToConcepts()`

src/apps/sources/__test__/redux/actions.test.ts
```diff
@@ -78,7 +78,7 @@ describe("action", () => {
     expect(indexedAction).toHaveBeenCalledWith("sources/retrieveSources", 1);
   });
 
-  it("for public sources should use action index as 2 along with the corresponding actionType and api", () => {
+  it("for All Public Concepts should use action index as 2 along with the corresponding actionType and api", () => {
```
ibacher, line 81: Don't replace this text... We only need things to read "All Public Concepts" in the UI or where we're testing something about the UI.

src/apps/sources/__test__/utils.test.ts
```diff
@@ -1,9 +1,9 @@
 import { getSourceTypeFromPreviousPath } from "../utils";
 
 describe("getSourceTypeFromPreviousPath", () => {
-  it("should return public sources for /sources/", () => {
+  it("should returnAll Public Concepts for /sources/", () => {
```
ibacher, line 4: Ditto on not needing "All Public Concepts" here

ibacher (changes requested):
I've added quite a few comments, including one that should fix the build error.

src/apps/sources/__test__/utils.test.ts
jwnasambu: @ibacher thanks for the review. Let me fix the proposed changes.

hadijahkyampeire:
@jwnasambu why do we have two PRS for this work, can you please close one and maintain one?

jwnasambu:
@hadijahkyampeire sorry for the confusion caused! After shifting to a new pr I forgot to close the later. Hopefully its fixed now.

hadijahkyampeire:
@jwnasambu we can still see two PRs for this work, please close one that you are not working on and maintain one which we can review.

src/apps/concepts/components/EnhancedTableToolbar.tsx
```diff
@@ -70,7 +70,7 @@ export const EnhancedTableToolbar = (props: EnhancedTableToolbarProps) => {
       )}
       {numSelected > 0 ? (
         <>
-          {!showAddConcepts ? null : (
+          {showAddConcepts ? null : (
```
ibacher, line 73: I see what you're doing here, but this should actually still be:

```ts
{!showAddConcepts ? null : (
```

Because it's still possible to be viewing concepts without a dictionary to add them to

src/apps/dictionaries/redux/actions.ts
```diff
@@ -389,12 +390,26 @@ export const recursivelyAddConceptsToDictionaryAction = (
...
+  updateProgress
+  );
+
+    const groupedConcepts = groupBy(concepts, "source");
+    console.log(groupedConcepts)
+    // const referencesToAdd = await Object.entries(groupedConcepts).map(([source, concepts]) => recursivelyFetchToConcepts(
+    //  source,
+    //  concepts.map((concept: { id: any; }) => concept.id),
+    // updateProgress,1,
+    //  ));
```
ibacher, line 412: Alright so this is very confusing, but here's what I *think* the implementation should look like:

```ts
    const groupedConcepts = groupBy(concepts, "source_url");
    const referencesToAdd = flatten(
      await Promise.all(
        Object.entries(groupedConcepts).map(([source, concepts]) =>
          recursivelyFetchToConcepts(
            source,
            concepts.map((concept) => concept.id),
            updateProgress
          )
        )
      )
    );
```

The trick here is that by calling `map()` we're building an array of... whatever the return value is. In this case, it's the return value of the call to `recursivelyFetchToConcepts()` which actually returns a `Promise<string[]>` (a [Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) to return an array of strings at a later point).

In the currently working version, the `await` before the `recursivelyFetchToConcepts()` waits until the call to `recursivelyFetchToConcepts()` returns. Since we're returning an array of promises, (one for each call to `recursivelyFetchToConcepts()`, i.e., one for each source), we need to wait until all those promises are finished. `Promise.all()` allows us to do this, turning an array of Promises into a promise that resolves with an array of results. We can safely `await` that.

Finally, I used `lodash`'s `flatten` function to transform the return time from `string[][]` (an array of arrays of strings) to a single array, that should look much like the one that existed before all these changes.

src/apps/dictionaries/redux/actions.ts
```diff
@@ -354,10 +355,10 @@ export const editDictionaryVersionAction = createActionThunk(
...
 export const recursivelyAddConceptsToDictionaryAction = (
-  sourceUrl: string,
   dictionaryUrl: string,
   rawConcepts: (APIConcept | string)[],
-  bulk: boolean = false
+  bulk: boolean = false,
+  sourceUrl?: string
 ) => {
   return async (dispatch: Function, getState: Function) => {
     const concepts = rawConcepts.map(concept =>
```
ibacher, line 364: To address my previous concern, this should look like this:

```ts
    if (!!!sourceUrl && !!rawConcepts.find((c) => typeof c === "string")) {
      throw {
        message: "Cannot load string-only concepts without a source url",
      };
    }

    const concepts = rawConcepts.map((concept) =>
      typeof concept === "string"
        ? {
            id: concept,
            url: `${sourceUrl}concepts/${concept}/`,
            source_url: sourceUrl,
          }
        : concept
    );
```

Note that this throws an error if `rawConcepts` contains any strings an no sourceUrl was provided. It also adds the `source_url` property to our concepts so we can use this in the `groupBy` below.

ibacher (commented):
@jwnasambu I've added some comments on a few points here. I'm not entirely sure this is all the changes necessary, but please try them out an let me know.

src/apps/dictionaries/redux/actions.ts
```diff
@@ -485,3 +483,11 @@ export const removeReferencesFromDictionaryAction = createActionThunk(
   REMOVE_REFERENCES_FROM_DICTIONARY,
   api.references.delete
 );
+function getState(): import("../../../redux").AppState {
+  throw new Error("Function not implemented.");
+}
+
+function dispatch(arg0: { type: string; actionIndex: number; meta: any[]; }) {
+  throw new Error("Function not implemented.");
+}
```
ibacher, line 492: Remove these... see my note above.

src/apps/concepts/components/EnhancedTableToolbar.tsx
```diff
@@ -70,7 +70,7 @@ export const EnhancedTableToolbar = (props: EnhancedTableToolbarProps) => {
       )}
       {numSelected > 0 ? (
         <>
-          {showAddConcepts ? null : (
+          {!showAddConcepts  ? null : (
```
ibacher, line 73: This line should read:

```ts
{showAddConcepts ? null : (
```

We need to change the conditions under which `showAddConcepts` is true, not just negate it (if we keep this in, this will show the `add concepts` part where we don't want to do that, i.e., where the user didn't start from their own dictionary).

src/apps/dictionaries/redux/actions.ts
```diff
@@ -357,20 +355,27 @@ export const editDictionaryVersionAction = createActionThunk(
...
-  return async (dispatch: Function, getState: Function) => {
-    const concepts = rawConcepts.map(concept =>
-      typeof concept === "string"
-        ? {
-            id: concept,
-            url: `${sourceUrl}concepts/${concept}/`
-          }
-        : concept
-    );
+  return async (dispatch: Function, getState: function) => {
```
ibacher, line 363: This is the small typo it's complaining about:

```suggestion
  return async (dispatch: Function, getState: Function) => {
```

`function` is a keyword `Function` is a type, if that makes any sense an here (after the `:`) this is serving as a type annotation.

ibacher (approved):
LGTM! Thanks @jwnasambu!

jwnasambu:
@ibacher thanks for your guidance you gave me a glimpse on how to work on a high level ticket and think out of the box.

PR #713, opened 2021-07-02
OCLOMRS-1010: Creating a public dictionary should create a public source

ibacher:
Also adds Cypress tests to confirm that the feature behaves as expected

# JIRA TICKET NAME:
[When creating a public dictionary, the source should also be public](https://issues.openmrs.org/browse/OCLOMRS-1010)

# Summary:

When a public dictionary is created, the source is also public. When a private dictionary is created, the source is private. When the dictionary visibility changes the source visibility changes to match.

### DIFF
--- a/cypress/integration/dictionary/create.feature
+++ b/cypress/integration/dictionary/create.feature
@@ -1,0 +1,38 @@
+Feature: Creating a dictionary
+  Background:
+    Given the user is logged in
+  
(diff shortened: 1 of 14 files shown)


hadijahkyampeire (approved):
Great work @ibacher with the dictionary tests and the reusable commands, also the public access works perfectly.

PR #723, opened 2021-07-26
OCLOMRS-913:The Basic User: AddConceptInDictionaryPage: Add a few concepts from CIEL in existing dictionary

jwnasambu:
# JIRA TICKET NAME:
[The Basic User: AddConceptInDictionaryPage: Add a few concepts from CIEL in existing dictionary](https://issues.openmrs.org/browse/OCLOMRS-913)

# Summary:


cypress/support/step_definitions/dictionary/addConceptInDictionaryPage/addconceptInDictioanryPage.ts
```diff
@@ -0,0 +1,69 @@
...
+  When
+} from "cypress-cucumber-preprocessor/steps";
+import { getDictionaryId, getUser, getConceptId } from "../../../utils";
+
+
+Given("the user is on the view dictionary concepts page", () => {
+  cy.visit(`/users/${getUser()}/collections/${getDictionaryId()}/concepts/`);
+});
+
+When("the user clicks the Add concepts button", () => {
```
ibacher, line 15: So, when I look at [the build log for this](https://github.com/openmrs/openmrs-ocl-client/pull/723/checks?check_run_id=3162725724#step:7:78), I see that the reason it is failing is this: `Error: Step implementation missing for: the user clicks the "Add concepts" button`. There are a couple of others like that.

In the `feature` file, [the step is written as](https://github.com/openmrs/openmrs-ocl-client/pull/723/files#diff-abfe500ea618d5b5cbed0ed71deb0289f2148580d01447a2d543eeff2bb87425R8): `When the user clicks the "Add concepts" button`. Note the quotation marks around `Add concepts` those need to be here two, so you can either write this as:

```ts
When('the user clicks the "Add concepts" button')
```

Or:

```ts
When("the user clicks the \"Add concepts\" button")
```

To get the same result. Similarly with other tests where you have quotation marks.

cypress/support/step_definitions/dictionary/addConceptInDictionaryPage/addconceptInDictioanryPage.ts
```diff
@@ -0,0 +1,69 @@
...
+  cy.findByTitle("Add concepts").click();
+});
+
+Then("the user should be on the Import existing concept page", () => {
+  cy.visit("/orgs/CIEL/sources/CIEL/concepts/?addToDictionary=/users/openmrs/collections/")
```
ibacher, line 20: This should be something like `cy.url().should("contain", "<url>")`. `cy.url()` gets the current URL. `cy.visit()` redirects the user to the provided url.

cypress/support/step_definitions/dictionary/addConceptInDictionaryPage/addconceptInDictioanryPage.ts
```diff
@@ -0,0 +1,69 @@
...
+});
+
+Given("the user is on the Import existing concept page", () => {
+  cy.visit("/orgs/CIEL/sources/CIEL/concepts/?addToDictionary")
```
ibacher, line 24: We should fill in the `addToDictionary` here, e.g.,

```suggestion
  cy.visit(`/orgs/CIEL/sources/CIEL/concepts/?addToDictionary=/users/${getUser()}/collections/${getDictionaryId()}/`)
```

cypress/support/step_definitions/dictionary/addConceptInDictionaryPage/addconceptInDictioanryPage.ts
ibacher: Also, instead of hard-coding `openmrs` as the user here, we should use `${getUser()}`. Note that to support this the string should be surrounded by back-ticks `` ` ``. (The difference between normal strings defined with `'` or `"` and strings defined with `` ` `` is that the strings defined with `` ` `` allow this kind of inline string interpolation.

cypress/support/step_definitions/dictionary/addConceptInDictionaryPage/addconceptInDictioanryPage.ts
```diff
@@ -0,0 +1,69 @@
...
+When("the user clicks on the row for Serum", () => {
+  cy.findByTitle("Serum").click();
+});
+
+When("the user clicks the Add selected to dictionary button", () => {
+  cy.findByTitle("Add selected to dictionary").click();
+});
+
+Then("the Serum concept should be added to the dictionary", () => {
+  cy.visit("/orgs/CIEL/sources/concepts/?addToDictionary=%2Fusers%2Fopenmrs%2Fcollections%2Fjst12%2F&page=1&q=Serum/")
```
ibacher, line 36: Here we should actually be testing the dictionary contents right?

ibacher (commented):
@jwnasambu I've added some comments here. Thanks for your hard work on this. I hope these comments are enough to get you unblocked!

jwnasambu:
@ibacher thanks for the review and the proposed changes.

cypress/support/step_definitions/dictionary/addConceptInDictionaryPage/addconceptInDictioanryPage.ts
```diff
@@ -0,0 +1,69 @@
...
+When('the user clicks on the row for "Serum"', () => {
+  cy.findByText("Serum").click();
+});
+
+When('the user clicks the "Add selected to dictionary" button', () => {
+  cy.findByTitle("Add selected to dictionary").click();
+});
+
+Then('the "Serum" concept should be added to the dictionary', () => {
+  cy.visit(`/orgs/CIEL/sources/CIEL/concepts/?addToDictionary=/users/${getUser()}/collections/${getDictionaryId()}/Serum/`)
```
jayasanka-sack, line 36: Isn't it should be something like this?

```suggestion
  cy.url().should('include', `/orgs/CIEL/sources/CIEL/concepts/?addToDictionary=/users/${getUser()}/collections/${getDictionaryId()}/Serum/`)
```

cypress/support/step_definitions/dictionary/addConceptInDictionaryPage/addconceptInDictioanryPage.ts
```diff
@@ -0,0 +1,69 @@
...
+When('the user clicks the "Add concepts" button', () => {
+  cy.findByTitle("Add concepts").click();
+});
+
+Then('the user should be on the "Import existing concept" page', () => {
+  cy.url().should("contain", "/orgs/CIEL/sources/CIEL/concepts/");
+});
+
+Given('the user is on the "Import existing concept" page', () => {
+  cy.url().should("contain", "/orgs/CIEL/sources/CIEL/concepts/");
```
jayasanka-sack, line 24: I guess this one should be
```suggestion
  cy.visit("/orgs/CIEL/sources/CIEL/concepts/");
```

cypress/support/step_definitions/dictionary/addConceptInDictionaryPage/addconceptInDictioanryPage.ts
jayasanka-sack: Check the following:
https://docs.cypress.io/api/commands/url#Assert-the-URL-is-http-localhost-8000-users-1-edit

jayasanka-sack: Is there any other way to determine whether the concept has been added? instead of checking the URL? A user-specific way.

ex: How a user knows when it has been created? Is it displays a success message? or it displays the created concept?

jwnasambu: Since for this case am importing the concept from the already existing concepts, it just displays the imported concept in the dictionary.

cypress/integration/dictionary/addConceptInDictionaryPage.feature
```diff
@@ -0,0 +1,42 @@
+Feature: Add concepts to an existing dictioanry
+  Background:
+    Given the user is logged in
+
+  @dictionary
+  Scenario: The user should be able to add concepts from their preferred source
+    Given the user is on the view dictionary concepts page
+     When the user clicks the "Add concepts" button
+      And the user selects "Import existing concept"
+      And the user selects "Pick concepts"
```
jayasanka-sack, line 10: Seems the step implementation missing for this one

cypress/support/step_definitions/dictionary/addConceptInDictionaryPage/addconceptInDictioanryPage.ts
```diff
@@ -0,0 +1,74 @@
...
+Then('the user should be on the "Import existing concept" page', () => {
+  cy.url().should("contain", "/orgs/CIEL/sources/CIEL/concepts/");
+});
+
+Given('the user is on the "Import existing concept" page', () => {
+  cy.visit(`/orgs/CIEL/sources/CIEL/concepts/?addToDictionary=/users/${getUser()}/collections/${getDictionaryId()}/`);
+});
+
+When('the user clicks on the row for "Serum"', () => {
+  cy.findByText("Serum").click();
```
jayasanka-sack, line 32: cypress throws the following error, 

```
Unable to find an element with the text: Serum. This could be because the text is broken up by multiple elements. In this case, you can provide a function for your text matcher to make your matcher more flexible.
```
as described in the error message, it seems there's no such element with the text "Serum"

cypress/support/step_definitions/dictionary/addConceptInDictionaryPage/addconceptInDictioanryPage.ts
```diff
@@ -0,0 +1,74 @@
...
+
+Given("the user is on the view dictionary concepts page", () => {
+  cy.visit(`/users/${getUser()}/collections/${getDictionaryId()}/concepts/`);
+});
+
+When('the user clicks the "Add concepts" button', () => {
+  cy.findByTitle("Add concepts").click();
+});
+
+Then('the user selects "Import existing concept"', () => {
```
jayasanka-sack, line 19: This on should be `When` not `Then`
```suggestion
When('the user selects "Import existing concept"', () => {
```

cypress/integration/dictionary/addConceptInDictionaryPage.feature
```diff
@@ -0,0 +1,42 @@
+Feature: Add concepts to an existing dictioanry
+  Background:
+    Given the user is logged in
```
jayasanka-sack, line 3: I couldn't find the step implementation for this one.

ibacher: We have a small number of steps that can be used anywhere because they would have basically the same definition everywhere. In this case, the implementation is [here](https://github.com/openmrs/openmrs-ocl-client/blob/72e2dfe39c12a5d6f3ec4dd131d5611a14713a41/cypress/support/step_definitions/common/common.ts#L9).

cypress/support/step_definitions/dictionary/addConceptInDictionaryPage/addconceptInDictioanryPage.ts
ibacher: Yes, to address this, we actually need to create a concept called "Serum" in a source called "CIEL" owned by an organisation called "CIEL". These tests are run against an empty instance of OCL (i.e., OCL without any content).

This will probably require a new command (since the `createSource` command only supports created a user source).

So we need something called before this function that does something like this:

```ts
cy.createOrganisation("CIEL", true).createOrgSource("CIEL", "CIEL", true);
```

And then afterwards,

```ts
cy.deleteOrganisation("CIEL", true);
```

Will also need a `createConcept` command that takes the source that the concept belongs to and the minimum number of properties we need so we can create a "Serum" concept before this test. Does that make some kind of sense?

ibacher:
@jwnasambu Please see @jayasanka-sack's helpful comments and let me know if there's any additional guidance you need.

jwnasambu:
@ibacher and @jayasanka-sack for the guidance. Ian, am soon reaching out.

jwnasambu:
@ibacher I kindly need your help help here! Though still building but am so confident it's going to fail.

jwnasambu:
@ibacher am glad the error on my terminal is the same as the one on the PR.

ibacher:
> am glad the error on my terminal is the same as the one on the PR.

That's actually very good news

ibacher:
@jwnasambu Hopefully at least three of these errors are straight-forward to fix:

* Step implementation missing for: the user clicks on the "Add Serum to dictionary" button
* Step implementation missing for: the user clicks the "Add selected to dictionary" button

Just need actual implementations and then (hopefully) those tests will pass.

jwnasambu:
@ibacher Thanks so much for the review though facing a different error on my terminal https://pastebin.com/QDki739E contrary to what is reflected on the PR. Basing on my step definition  all the implementations exist as reflected here https://pastebin.com/bsCNKYPz.

jayasanka-sack:
@jwnasambu The element you trying to click has CSS `pointer-events: none`. That means the element is not clickable. 

Try clicking on a clickable element, or you can user `{force: true}` to click it forcefully (not a good idea, btw).

suruchee:
@jwnasambu, will you please update the status of this PR? And if everything is fixed, will you please remove it from draft?

jwnasambu:
@suruchee I will be the happiest person to change the status this PR but us you can see the build is still failing and basing on Ian's explanation `Oh… I kind of see what the issue is. The loading of CIEL concepts is dependent on them not existing (since, in the environment that we run them in for CI, they don’t exist), but if you run it against staging, as you’re doing here, they’ll fail because CIEL already exists and has those concepts (and the user we login with doesn’t have permission to create CIEL concepts).
The way to fix that is probably just to check if the CIEL Org, Source and concepts exist and only try to create them if they don’t. We have similar checks like that for some of the other create commands. I can work on that tomorrow, but I’m kind of busy the rest of this evening.` Once I fix the issue, the PR status will definitely change. Thanks in advance.

jwnasambu:
@jayasanka-sack am sorry I had not your comment. Kindly forgive me.

jwnasambu:
> @jwnasambu The element you trying to click has CSS `pointer-events: none`. That means the element is not clickable.
> 
> Try clicking on a clickable element, or you can use `{force: true}` to click it forcefully (not a good idea, btw).

Am still getting the same error.

cypress/support/step_definitions/dictionary/addConceptsToDictionary/addConceptsToDictionary.ts
```diff
@@ -0,0 +1,170 @@
...
+
+When('the user clicks on the link for "Serum"', () => {
+  cy.findByText("Serum").click();
+});
+
+When("the user is sent to the view concept page", () => {
+  cy.visit("users/openmrs/collections");
+});
+
+When('the user clicks on the "Add selected to dictionary" button', () => {
```
jayasanka-sack, line 112:
```suggestion
When('the user clicks the "Add selected to dictionary" button', () => {
```

cypress/support/step_definitions/dictionary/addConceptsToDictionary/addConceptsToDictionary.ts
```diff
@@ -0,0 +1,170 @@
...
+});
+
+When('the user clicks on the row for "Serum"', () => {
+  cy.findByText("Serum")
+    .parent()
+    .next()
+    .click();
+});
+
+When('the user clicks the "Add Serum to dictionary" button', () => {
```
jayasanka-sack, line 59:
```suggestion
When('the user clicks on the "Add Serum to dictionary" button', () => {
```

jwnasambu: @jayasanka-sack and @ibacher  it has be a tough journey but am glad am remaining with 1 error though the error on the PR is different from the one on the terminal.

ibacher (approved):

ibacher:
@hadijahkyampeire If you have a chance, could you take a look at the changes [to the ViewConceptsPage.tsx](https://github.com/openmrs/openmrs-ocl-client/pull/723/files#diff-60efd061211b613f4964e3afe828937492f8764f1f49790b47eb8cb19a778bbb) and let me know if any of the changes there don't make sense?

jwnasambu:
@ibacher and @jayasanka-sack I can't thank you enough. It has been a real fight I now have a glimpse of what it means to be a senior dev. Though with serious errors!

hadijahkyampeire:
> @hadijahkyampeire If you have a chance, could you take a look at the changes [to the ViewConceptsPage.tsx](https://github.com/openmrs/openmrs-ocl-client/pull/723/files#diff-60efd061211b613f4964e3afe828937492f8764f1f49790b47eb8cb19a778bbb) and let me know if any of the changes there don't make sense?

Okay @ibacher I will do that.

hadijahkyampeire (approved):
I think all is fine, they all make sense and will make the importing from another dictionary work well, great work @ibacher  and @jwnasambu . LGTM

PR #724, opened 2021-07-26
OCLOMRS-970:Automated test for creating a version, releasing it and copy subscriptionURL

suruchee:
# JIRA TICKET NAME:
[The Loader-Upper](https://issues.openmrs.org/browse/OCLOMRS-970)

# Summary:
Automated test for creating a dictionary version, releasing it, and copy subscription URL


cypress/integration/dictionary/createVersion.feature
```diff
@@ -0,0 +1,31 @@
...
+      And the user clicks yes button
+    Then the version should be released
+    When the user clicks the more actions button
+      And the user selects the "Copy Subscription URL" menu list item
+    Then the subscription url should be copied
+
+
+
+
+
```
ibacher, line 31:
```suggestion

```

cypress/support/step_definitions/common/common.ts
```diff
@@ -19,10 +26,12 @@ Given("a dictionary exists", () => {
 Before({ tags: "@dictionary" }, () => {
   setDictionaryId(`TD-${nanoid()}`);
 });
-
 Before({ tags: "@concept" }, () => {
   setConceptId(`CT-${nanoid()}`);
 });
+Before({ tags: "@version" }, () => {
```
ibacher, line 32:
```suggestion

Before({ tags: "@version" }, () => {
```

cypress/support/step_definitions/dictionary/createVersion/createVersion.ts
```diff
@@ -0,0 +1,57 @@
...
+
+Given("the user is on the dictionary page", () => {
+  cy.visit(`/users/${getUser()}/collections/${getDictionaryId()}/`);
+  cy.findByText("Versions").should("be.visible");
+});
+When("the user clicks the create new version button", () => {
+  cy.findByRole("button", { name: /Create new version/i }).click();
+});
+Then("the user should be on the create new version dialog box", () =>
+  cy.get("h2").contains("Create new version")
```
ibacher, line 12:
```suggestion
  cy.findByText("Create new version").should("be.visible")
```

cypress/support/step_definitions/dictionary/createVersion/createVersion.ts
```diff
@@ -0,0 +1,57 @@
...
+);
+
+Given("the user is on the create new version dialog box", () => {
```
ibacher, line 15:
```suggestion
Given("the user clicks on the create new version dialog box", () => {
```

And a similar change to the `feature` file.

cypress/support/step_definitions/dictionary/createVersion/createVersion.ts
```diff
@@ -0,0 +1,57 @@
...
+  cy.visit(`/users/${getUser()}/collections/${getDictionaryId()}/`);
+  cy.findByRole("button", { name: /Create new version/i }).click();
+  cy.get("h2").contains("Create new version");
+});
+When("the user enters the version information", () => {
+  cy.findByLabelText("ID").type("1");
+  cy.get("#released").type("{enter}");
+});
+When("the user submits the form", () => {
+  cy.get("#versionForm").submit();
```
ibacher, line 25: There's no `#versionForm` element, so just do:

```ts
cy.findByRole("button", { name: /Submit/i }).click();
```

cypress/support/step_definitions/dictionary/createVersion/createVersion.ts
```diff
@@ -0,0 +1,57 @@
+import { Given, Then, When } from "cypress-cucumber-preprocessor/steps";
+import {getDictionaryId, getUser, getVersionId} from "../../../utils";
+
+Given("the user is on the dictionary page", () => {
+  cy.visit(`/users/${getUser()}/collections/${getDictionaryId()}/`);
+  cy.findByText("Versions").should("be.visible");
+});
+When("the user clicks the create new version button", () => {
+  cy.findByRole("button", { name: /Create new version/i }).click();
+});
```
ibacher, line 10:
```suggestion
When("the user clicks the create new version button", () =>
  cy.findByRole("button", { name: /Create new version/i }).click()
);
```

cypress/support/step_definitions/dictionary/createVersion/createVersion.ts
```diff
@@ -0,0 +1,57 @@
...
+When("the user submits the form", () => {
+  cy.get("#versionForm").submit();
+});
+Then("the new version should be created", () => {
+  cy.get("table").should("exist");
+  cy.get("table tbody tr").should("have.length", 1);
+  cy.get("table").contains("td", "1");
+  cy.get("table").contains("td", "No");
+});
```
ibacher, line 32: There are probably too many assertions here we don't need. E.g., what we care about is there is a row with the ID we filled in. Also, I'm not sure what the "No" here is supposed to match, but we'd probably need to check the actual checkbox if its the release status.

cypress/integration/dictionary/createVersion.feature
```diff
@@ -0,0 +1,31 @@
...
+  @dictionary
+  Scenario: The user should be able to click the button to create a new version
+    Given a dictionary exists
+      And the user is on the dictionary page
+    When the user clicks the create new version button
+    Then the user should be on the create new version dialog box
+
+  @dictionary
+  @version
+  Scenario: The user should be able to create a new version, release it and copy subscription URL
```
ibacher, line 14: I would actually break this out as three separate tests: one to create a new version. One to release it via the checkbox, and one to copy the subscription URL. That way we should also be able to check that the user can release a new version within the "Create New Version" form.

cypress/support/step_definitions/dictionary/createVersion/createVersion.ts
```diff
@@ -0,0 +1,57 @@
...
+
+When("the user clicks release status switch", () => {
+  cy.get('[type="checkbox"]').check();
+});
+When("the release dialog opens", () => {
+  cy.get("#confirmation-dialog-title").should("exist");
```
ibacher, line 38:
```suggestion
  cy.get("#confirmation-dialog-title").should("be.visible");
```

cypress/support/step_definitions/dictionary/createVersion/createVersion.ts
```diff
@@ -0,0 +1,57 @@
...
+});
+When("the user clicks yes button", () => {
+  cy.get("button.MuiButton-textPrimary")
+    .eq(1)
+    .click();
```
ibacher, line 43: cy.findByText(/Yes/i).click()

cypress/support/step_definitions/dictionary/createVersion/createVersion.ts
```diff
@@ -0,0 +1,57 @@
...
+);
+
+When("the user clicks the more actions button", () => {
+  cy.findByTitle("More actions").click();
+});
+When(/the user selects the "(.+)" menu list item/, menuItem =>
+  cy.findByText(menuItem).click()
+);
+Then("the subscription url should be copied", () =>
+cy.visit(`/users/${getUser()}/collections/${getDictionaryId()}/${getVersionId()}`)
```
ibacher, line 56: I'm not sure I see how this code checks that the subscription URL was copied? [This](https://github.com/cypress-io/cypress/issues/2752#issuecomment-734636958) looks like a good approach to verify that the correct text was copied.

cypress/support/utils.ts
```diff
@@ -26,7 +26,11 @@ export const getDictionaryId = () => Cypress.env("dictionaryId");
...
 export const getConceptId = () => Cypress.env("conceptId");
-export const setConceptId = (conceptId: string) => Cypress.env("conceptId", conceptId);
+export const setConceptId = (conceptId: string) =>
+  Cypress.env("conceptId", conceptId);
 export const getOrganisationId = () => Cypress.env("organisationId");
 export const setOrganisationId = (organisationId: string) =>
   Cypress.env("organisationId", organisationId);
+export const getVersionId = () => Cypress.env("versionId");
+export const setVersionId = (versionId: string) =>
+  Cypress.env("versionId", versionId);
```
ibacher, line 36:
```suggestion
  Cypress.env("versionId", versionId);

```

ibacher (commented):
Thanks for this @suruchee! I've added a bunch of comments on some things to tighten this up a bit, but very good work!

cypress/support/step_definitions/dictionary/createVersion/createVersion.ts
suruchee: Hi @ibacher, but this code will cause **Found multiple elements with the text** error. So I tried this one. Will you please suggest another idea?

suruchee:
Thank you @ibacher, for the review. I have updated the changes but I am not sure about making separate scenarios for release and copy url.

cypress/integration/dictionary/createVersion.feature
hadijahkyampeire: You are right @ibacher that's what @suruchee had but the last two scenarios were not passing probably because they needed a version to be created, she had a command for creating and getting a version the way the dictionary commands are but they were not working, so I told her that we first put those other two tests in the same scenario so that we see they pass, then we reach out to you about separating them.

hadijahkyampeire (approved):
Very well done @suruchee, I think this will be good to merge once we do the separation with @ibacher 's guidance.

cypress/support/step_definitions/dictionary/createVersion/createVersion.ts
ibacher:
```ts
cy.findByText("Create new version", { selector: "h2" }).should("be.visible")
```

cypress/integration/dictionary/createVersion.feature
ibacher: Fair enough... The version release process actually does some backend processing, so between the request to create the version returning and the version actually being created there's a bit of a delay... not much for a small dictionary, but not no time at all either.

Probably the simplest way to do this is to force Cypress to pause immediately after submitting the new version request, e.g., by using `cy.wait(500)`. A more complicated version of the same idea would be to have a shorter pause and at the end of each pause poll the backend until a longer period of time, e.g. 8 seconds or w/e had passed.

ibacher: Here's a library to handle my "more complicated" version: https://github.com/NoriSte/cypress-wait-until. Basically looks like you can use that and wait until the backend returns that the version exists. Note that it's probably also necessary to force Cypress to reload the page.

suruchee: Thank you @hadijahkyampeire  and @ibacher, I will update with the delay suggestion.

PR #729, opened 2021-08-17
OCLOMRS-749: The Basic User: AddBulkConcepts to the dictionary

jwnasambu:
# JIRA TICKET NAME:
[The Basic User: AddBulkConcepts to the dictionary](https://issues.openmrs.org/browse/OCLOMRS-749)

# Summary:
I created the feature file and step description for add bulk concepts to the dictionary.


cypress/integration/dictionary/AddBulkConceptsToDictionaryPage/addBulkConceptsToDictionary.feature
```diff
@@ -0,0 +1,32 @@
...
+    Given the user is on the "Add concepts in bulk from CIEL" page
+     When the user enters concept Id "1000"
+      And the user enters concept Id "1001"
+      And the user enters concept Id "1002"
+      And the user clicks the "Add concepts" button
+      And the user is sent to the "Progress notifications" page
+     Then the user clicks on "view summary" button
+       And concept Id "1000" should be skippped
+       And the concept Id "1001" should be added to the dictionary
+       And the concept Id "1002" should be added to the dictionary
```
ibacher, line 32: It's always a good idea to add a blank line to the end of Git-tracked files.

```suggestion
       And the concept Id "1002" should be added to the dictionary

```

cypress/integration/dictionary/AddBulkConceptsToDictionaryPage/addBulkConceptsToDictionary.feature
```diff
@@ -0,0 +1,32 @@
...
+  @dictionary
+  Scenario: The user should be able to add bulk concepts from their preferred source
+    Given the user is on the view dictionary concepts page
+     When the user clicks the "Add concepts" button
+      And the user selects "Import existing concept"
+      And the user selects "Add bulk concepts"
+     Then the user should be on the "Add concepts in bulk from CIEL" page
+
+  @dictionary
+  Scenario: The user should be able to add a single bulk concept from their preferred source
```
ibacher, line 14: Remeber how to support the `addConceptsToDictionary` feature we had to add [this](https://github.com/openmrs/openmrs-ocl-client/blob/570ca1cbe0b603814eaaf227f35772f56ba838b6/cypress/support/step_definitions/dictionary/addConceptsToDictionary/addConceptsToDictionary.ts#L108) to ensure that the CIEL dictionary exists and had some concepts in it? We need to do the same thing here. It might be a good idea to migrate those before and after hooks to [the common file](https://github.com/openmrs/openmrs-ocl-client/blob/master/cypress/support/step_definitions/common/common.ts).

cypress/integration/dictionary/AddBulkConceptsToDictionaryPage/addBulkConceptsToDictionary.feature
```diff
@@ -0,0 +1,32 @@
...
+  @dictionary
+  Scenario: The user should be able to add multiple bulk concepts from their preferred source
+    Given the user is on the "Add concepts in bulk from CIEL" page
+     When the user enters concept Id "1000"
+      And the user enters concept Id "1001"
+      And the user enters concept Id "1002"
+      And the user clicks the "Add concepts" button
+      And the user is sent to the "Progress notifications" page
+     Then the user clicks on "view summary" button
+       And concept Id "1000" should be skippped
```
ibacher, line 30: Why should "1000" be skipped? Remember that each scenario (should) run in isolation from other scenarios, so if we're expecting that concept 1000 is already in the dictionary, then we should have an explicit "Given" step to address this. 

```
Given CIEL concept "1000" is already in the dictionary
  And the user is on the "Add concepts in bulk from CIEL" page

File: cypress/integration/dictionary/AddBulkConceptsToDictionaryPage/addBulkConceptsToDictionary.feature (hunk 1 above)
Comment (jwnasambu) [reply cypress/integration/dictionary/AddBulkConceptsToDictionaryPage/addBulkConceptsToDictionary.feature:32]: @ibacher am sorry for a bother! Kindly I don't seem to understand this commend do you mind helping please?

Comment (ibacher) [reply cypress/integration/dictionary/AddBulkConceptsToDictionaryPage/addBulkConceptsToDictionary.feature:32]: I just mean that in your file you should have an empty line with no spaces or indentation, so the end of the file looks like this:

```
       And the concept Id "1002" should be added to the dictionary

```

Rather than:

```
       And the concept Id "1002" should be added to the dictionary
```

Comment (jwnasambu) [reply cypress/integration/dictionary/AddBulkConceptsToDictionaryPage/addBulkConceptsToDictionary.feature:32]: @ibacher Thanks for the clear explanation.

File: cypress/support/step_definitions/dictionary/addBulkConceptsToDictionary.ts [hunk 4]
```diff
@@ -0,0 +1,128 @@
...
+
+Given('CIEL concept "1000" is already in the dictionary', () => {
+  cy.visit('/actions/');
+  cy.url().should(
+    "contain", '/actions/'
+  );
+});
+
+When('the user clicks the "Add concepts" button', () => {
+  cy.findByTitle("Add concepts").click();
```
ibacher, line 37: So the problem here is that you're asking it to find the button by the `title` property with the value of "Add concepts", but in the [source code](https://github.com/openmrs/openmrs-ocl-client/blob/c3c7d5246006bd0619ee70e947ae2d59a992d392/src/apps/dictionaries/pages/AddBulkConceptsPage.tsx#L107), we don't set the `title` property. Consequently, in the rendered document, the mark-up to produce the document looks like this:

```
<div class="jss100">
	<a class="MuiButtonBase-root MuiButton-root MuiButton-outlined MuiButton-outlinedPrimary" tabindex="0" role="button" aria-disabled="false" href="/actions">
		  <span class="MuiButton-label">
			  Add concepts
		  </span>
		  <span class="MuiTouchRipple-root"></span>
	</a>
</div>
```

(I got this by right-clicking on the button and selecting "Inspect"). Instead, you'll want to use `findByText()` here since the "Add concepts" string is the text of the node we want to click on.

jwnasambu: @ibacher thanks for the review let me fix it right away.

cypress/integration/dictionary/addBulkConceptsToDictionary.feature
```diff
@@ -0,0 +1,44 @@
...
+       And the concept Id "1002" should be added to the dictionary 
+
+@dictionary
+@ciel
+  Scenario: The system should be able to skip an already added bulk concepts from CIEL
+    Given CIEL concept "1000" is already in the dictionary
+      And the user is on the "Add concepts in bulk from CIEL" page
+      And the user re-enters concept Id "1000"
+      And the user clicks the "Add concepts" button
+      And the user is sent to the "Progress notifications" page
```
jayasanka-sack, line 41: Step implementation is missing

cypress/integration/dictionary/addBulkConceptsToDictionary.feature
```diff
@@ -0,0 +1,44 @@
...
+     Then the user clicks on "view summary" button
+      And concept Id "1000" should be skippped 
+
```
jayasanka-sack, line 44: Add an empty newline at the end of the file

cypress/support/index.ts
```diff
@@ -17,9 +17,13 @@
...
+      // tslint:disable: no-console
+      console.log('error', err)
+      console.log('runnable', runnable)
+  }
   // returning false here prevents Cypress from
   // failing the test
-  return false;
-});
+  return false
+})
```
jayasanka-sack, line 29: Add an empty newline at the end of the file

cypress/support/step_definitions/dictionary/addBulkConceptsToDictionary/addBulkConceptsToDictionary.ts
```diff
@@ -0,0 +1,128 @@
...
+    cy.get("textarea")
+  }
+); 
+
+When('the user re-enters concept Id "1000"', () => {
+  cy.getConcept("#conceptId")
+  cy.get("textarea")
+});
+
+When("the user is send to progress notifications page", () => {
```
jayasanka-sack, line 58:
```suggestion
When('the user is send to "Progress notifications" page', () => {
```

cypress/support/step_definitions/dictionary/addBulkConceptsToDictionary/addBulkConceptsToDictionary.ts
```diff
@@ -0,0 +1,128 @@
...
+
+Given('CIEL concept "1000" is already in the dictionary', () => {
+  cy.visit('/actions/');
+  cy.url().should(
+    "contain", '/actions/'
+  );
+});
+
+When('the user clicks the "Add concepts" button', () => {
+  cy.findByText("Add concepts").click({force: true});
```
jayasanka-sack, line 37: Error from cypress:

`Unable to find an element with the text: Add concepts. This could be because the text is broken up by multiple elements. In this case, you can provide a function for your text matcher to make your matcher more flexible.`

jayasanka-sack (dismissed):
Good job @jwnasambu !

I have requested a set of changes. Please check them and let me know if you have any doubts.

cypress/integration/dictionary/addBulkConceptsToDictionary.feature
```diff
@@ -0,0 +1,44 @@
+Feature: Add bulk concepts to an existing dictionary
+  Background:
+    Given the user is logged in
+
+  @dictionary
+  Scenario: The user should be able to add bulk concepts from their preferred source
+    Given the user is on the view dictionary concepts page
+     When the user clicks the "Add concepts" button
```
ibacher, line 8: We now have two steps called "the user clicks the "Add concepts" button" trying to test very different things (one is clicking the "Add concepts" button on the bulk import page. One is clicking the little "+" icon on the main dictionary view.

We need to describe these steps differently so we can have different implementations for them.

cypress/support/step_definitions/dictionary/addBulkConceptsToDictionary/addBulkConceptsToDictionary.ts
ibacher: I wouldn't use `{force: true}` without very good reason. If the button is disabled and we're forcing the click on it, we're doing something the end-user can't. If the button isn't enabled yet, we need to figure out why.

cypress/support/step_definitions/dictionary/addBulkConceptsToDictionary/addBulkConceptsToDictionary.ts
```diff
@@ -0,0 +1,128 @@
...
+});
+
+When('the user selects "Add bulk concepts"', () => {
+cy.visit(
+  `/users/${getUser()}/collections/${getDictionaryId()}/add/?fromSource=CIEL/`
+  );
```
ibacher, line 43:
```suggestion
  cy.visit(
    `/users/${getUser()}/collections/${getDictionaryId()}/add/?fromSource=CIEL/`
  );
```

ibacher: This isn't actually doing what's described in the text, though. In the text, we're supposed to be finding the element labelled "Add bulk concepts" and click on it to be redirected to the location... this just redirects directly there.

cypress/support/step_definitions/dictionary/addBulkConceptsToDictionary/addBulkConceptsToDictionary.ts
ibacher: "send" -> "sent". However, this should probably be "the user navigates to the progress notifications page" or similar. Note, too, that there's likely to be a delay. Rather than jumping over to another screen, it might be better just to verify that the expected concepts are added to the dictionary. We can have some separate tests for what ought to be on the notification screen.

cypress/support/step_definitions/dictionary/addBulkConceptsToDictionary/addBulkConceptsToDictionary.ts
jwnasambu: @ibacher and @jayasanka-sack thanks for your review and proposed changes. Let me fix them right away!

cypress/integration/dictionary/addBulkConceptsToDictionary.feature
jwnasambu: @ibacher thanks so so much I deleted  the little "+" icon on the main thinking I have repeated the implementation. Thanks so so much for helping me spot my endless errors.

cypress/support/step_definitions/dictionary/addBulkConceptsToDictionary/addBulkConceptsToDictionary.ts
jwnasambu: @ibacher the explanation makes sense to me. Thanks so so much!

cypress/support/step_definitions/dictionary/addBulkConceptsToDictionary/addBulkConceptsToDictionary.ts
```diff
@@ -0,0 +1,132 @@
...
+
+Given('CIEL concept "1000" is already in the dictionary', () => {
+  cy.visit('/actions/');
+  cy.url().should(
+    "contain", '/actions/'
+  );
+});
+
+When('the user clicks the "Add concepts" button', () => {
+  cy.findByTitle("Add concepts").click();
```
jayasanka-sack, line 37: maybe `cy.contains("Add concepts").click();` ?

cypress/integration/dictionary/addBulkConceptsToDictionary.feature
ibacher: We still need to be able to distinguish between "the user clicks the "Add concepts" button" here and the other one we use. The element that we display the "+" with is called a ["floating action button"](https://material.io/components/buttons-floating-action-button) in the Material Design framework because it's designed to remain on the user's screen regardless of scroll position, height, etc. I.e., it's "floating" there in the lower-right.

So let's name this step:

```
When the user clicks on the "Add concepts" floating action button
```

And leave the other text for the other button with an "Add concepts" label.

cypress/support/step_definitions/common/common.ts
```diff
@@ -73,3 +73,49 @@ After({ tags: "@organisation" }, () => {
...
+      ],
+      "/orgs/CIEL/sources/CIEL/",
+      "1002"
+    );
+});
+
+After({ tags: "@ciel" }, () => {
+ cy.deleteOrgSource("CIEL", "CIEL", true).deleteOrganisation("CIEL", true);
+});
+
```
ibacher, line 121: For now, let's not change common.ts. We can do this as a new ticket at some point.

cypress/support/step_definitions/dictionary/addBulkConceptsToDictionary/addBulkConceptsToDictionary.ts
jwnasambu: @jayasanka-sack I still end up with error `AssertionError
Timed out retrying after 8000ms: Expected to find content: 'Add concepts' but never did.`  when I use `cy.contains("Add concepts").click();`

cypress/support/step_definitions/common/common.ts
jwnasambu: @ibacher I have removed the proposed content

cypress/support/step_definitions/dictionary/addBulkConceptsToDictionary/addBulkConceptsToDictionary.ts
jwnasambu: > This isn't actually doing what's described in the text, though. In the text, we're supposed to be finding the element labelled "Add bulk concepts" and click on it to be redirected to the location... this just redirects directly there.

@ibacher am sorry for a bother! did I miss out on the location?

cypress/integration/dictionary/addBulkConceptsToDictionary.feature
jwnasambu: @ibacher am sorry I have just seen the above comment but I have fixed it.

cypress/support/step_definitions/dictionary/addBulkConceptsToDictionary/addBulkConceptsToDictionary.ts
```diff
@@ -0,0 +1,130 @@
...
+
+When('the user selects "Add bulk concepts"', () => {
+  cy.visit(
+    `/users/${getUser()}/collections/${getDictionaryId()}/add/?fromSource=CIEL/`
+  );
+});
+
+When(
+  /the user enters concept Id "(1000|1001|1002)"/, (conceptId) => {
+    cy.get("textarea").type("conceptId")
```
ibacher, line 52: We want to use the variable `conceptId` here, not the string "conceptId"

```suggestion
    cy.get("textarea").type(conceptId)
```

cypress/support/step_definitions/dictionary/addBulkConceptsToDictionary/addBulkConceptsToDictionary.ts
```diff
@@ -0,0 +1,130 @@
...
+  }
+); 
+
+When('the user re-enters concept Id "1000"', (conceptId) => {
+  cy.get("textarea").type("conceptId")
+});
+
+When('the user navigates to the "Progress notifications" page', () => {
+   cy.url().should("actions/");
```
ibacher, line 61: Please see the [syntax for Cypress's `should()` function](https://docs.cypress.io/api/commands/should). It needs both a "chainer", that is an assertion, and a value. E.g.,

```js
cy.url().should("equal", "actions/")
```

cypress/support/step_definitions/dictionary/addBulkConceptsToDictionary/addBulkConceptsToDictionary.ts
```diff
@@ -0,0 +1,130 @@
...
+});
+
+Then('the user should be on the "Add concepts in bulk from CIEL" page', () => {
+   cy.url().should("contain", `/users/${getUser()}/collections/${getDictionaryId()}/add/?fromSource=CIEL/`);
+});
+
+Then(
+  /the "(1000|1001|1002)" concepts should be added to the dictionary/, conceptId => {
+    cy.getConcept(`/users/${getUser()}/collections/${getDictionaryId()}/concepts/${conceptId}`).should("exist");
+    cy.get("textarea").type("conceptId")
```
ibacher, line 75: We don't need this line here. There's no text area. More generally `When()` steps describe an action the user does and `Then()` steps describe an assertion.

cypress/support/step_definitions/dictionary/addBulkConceptsToDictionary/addBulkConceptsToDictionary.ts
```diff
@@ -0,0 +1,130 @@
...
+  });    
+
+Then('the "1000" concept should be skippped', conceptId => {
+  cy.getConcept(`/users/${getUser()}/collections/${getDictionaryId()}/concepts/${conceptId}`).should("skip");
```
ibacher, line 79: `should()` isn't going to understand what "skip" means. (`"exist"` is [a defined assertion](https://docs.cypress.io/guides/references/assertions#Chai)).

More generally, we don't need this step in the validation. What we are validating is that the concept was skipped, which we aren't going to learn from loading the concept from the backend (which is what `cy.getConcept()` does).

cypress/integration/dictionary/addBulkConceptsToDictionary.feature
```diff
@@ -0,0 +1,45 @@
...
+  Scenario: The system should be able to skip an already added bulk concepts from CIEL
+    Given CIEL concept "1000" is already in the dictionary
+      And the user is on the "Add concepts in bulk from CIEL" page
+      And the user re-enters concept Id "1000"
+      And the user clicks the "Add concepts" MuiButton-label button
+      And the user navigates to the "Progress notifications" page
+     Then the user clicks on "view summary" button
+      And concept Id "1000" should be skippped 
+      
+
```
ibacher, line 45:
```suggestion

```

cypress/support/index.ts
ibacher, line 29:
```suggestion
  return false;
});
```

ibacher (changes requested):

cypress/integration/dictionary/addBulkConceptsToDictionary.feature
jwnasambu: @ibacher I don't seem to understand this comment.

cypress/support/step_definitions/dictionary/addBulkConceptsToDictionary/addBulkConceptsToDictionary.ts
jwnasambu: @ibacher does it mean I delete this step?

cypress/integration/dictionary/addBulkConceptsToDictionary.feature
ibacher: @jwnasambu You've got two lines with nothing but spaces in them at the end of the file. I was sugggesting deleting them and leaving a single blank line with no content at the end of the file.

jwnasambu:
@ibacher kindly do you mind helping me on this error please? I missed out on something which am an able to figure out..

cypress/support/step_definitions/dictionary/addBulkConceptsToDictionary/addBulkConceptsToDictionary.ts
jwnasambu: Concerning this error `CypressError: `cy.type()` can only accept a string or number. You passed in: `undefined`` I thought I was to use a string in `.type("conceptId")` Kindly what is your take?

cypress/support/step_definitions/dictionary/addBulkConceptsToDictionary/addBulkConceptsToDictionary.ts
ibacher: Yes

cypress/support/step_definitions/dictionary/addBulkConceptsToDictionary/addBulkConceptsToDictionary.ts
ibacher: My point was that we shouldn't be calling `cy.type()` here at all.

cypress/support/step_definitions/dictionary/addBulkConceptsToDictionary/addBulkConceptsToDictionary.ts
```diff
@@ -0,0 +1,130 @@
...
+  );
+});
+
+When(
+  /the user enters concept Id "(1000|1001|1002)"/, (conceptId) => {
+    cy.get("textarea").type(conceptId)
+  }
+); 
+
+When('the user re-enters concept Id "1000"', (conceptId) => {
```
ibacher, line 56: What is the value of `conceptId` here?

ibacher: It may be helpful to look at other examples where we use an argument like `conceptId`

cypress/support/step_definitions/dictionary/addBulkConceptsToDictionary/addBulkConceptsToDictionary.ts
```diff
@@ -0,0 +1,130 @@
...
+  cy.get("textarea").type(conceptId)
+});
+
+When('the user navigates to the "Progress notifications" page', () => {
+  cy.url().should("equal", "actions/")
+});
```
ibacher, line 62: A couple of things here:

1. The text "the user navigates to the "Progress notifications" page" sort of implies that there's a call to `cy.visit()` here.
2. `equal` may not be the right [assertion](https://docs.cypress.io/guides/references/assertions#Chai) to use here.

cypress/support/step_definitions/dictionary/addBulkConceptsToDictionary/addBulkConceptsToDictionary.ts
```diff
@@ -0,0 +1,130 @@
...
+  );
+});
+
+Given('the user is on the "Add concepts in bulk from CIEL" page', () => {
+cy.visit(
+`/users/${getUser()}/collections/${getDictionaryId()}/add/?fromSource=CIEL/`
+);
+cy.url().should(
+"contain",
+`/users/${getUser()}/collections/${getDictionaryId()}/add/?fromSource=CIEL/`
```
ibacher, line 25:
```suggestion
  cy.visit(
    `/users/${getUser()}/collections/${getDictionaryId()}/add/?fromSource=CIEL/`
  );
  
  cy.url().should(
    "contain",
    `/users/${getUser()}/collections/${getDictionaryId()}/add/?fromSource=CIEL/`
```

cypress/support/step_definitions/dictionary/addBulkConceptsToDictionary/addBulkConceptsToDictionary.ts
```diff
@@ -0,0 +1,130 @@
...
+  );
+});
+
+Given('CIEL concept "1000" is already in the dictionary', () => {
+  cy.visit('/actions/');
+  cy.url().should(
+    "contain", '/actions/'
+  );
+});
```
ibacher, line 34: What does this step do?

ibacher (changes requested):
@jwnasambu

In the error log there's [this message](https://github.com/openmrs/openmrs-ocl-client/runs/3614555864#step:7:79):

```
      + expected - actual

      -'http://localhost:8080/actions'
      +'actions/'
```

This indicates that there's some point where you've asked Cypress to look for "actions/" (the `+` means expected) but the actual value is `http://localhost:8080/actions`. So those checks should be changed to something that will match the actual value (assuming that `http://localhost:8080/actions` is the correct result).

[This error message](https://github.com/openmrs/openmrs-ocl-client/runs/3614555864#step:7:104) is pretty self-explanatory:

```
     CypressError: `cy.type()` can only accept a string or number. You passed in: `undefined`
```

Somewhere you have a call to `cy.type()` where the value that you pass to it evaluates to `undefined`. Most likely, this is something like `cy.type(variable)` where the value of the `variable` hasn't been set for whatever reason.

cypress/support/step_definitions/dictionary/addBulkConceptsToDictionary/addBulkConceptsToDictionary.ts
jwnasambu: @ibacher it is supposed to reflect the page [actions](http://localhost:8080/actions) where the concept Id is already added.

ibacher: @jwnasambu Surely with a step like this we want to check that the concept already exists in the dictionary rather than to go to the progress notifications page right? This is a "Given" step, so it should be asserting some kind of precondition, e.g.,

```ts
cy.getConcept(`/users/${getUser()}/collections/${getDictionaryId()}/concepts/1000`).should("exist");
```

jwnasambu:
@ibacher Kindly forgive me if I missed out on any proposed changes.

jwnasambu:
@ibacher I must admit I need help am confusing myself more and more with the many error that am getting.

jwnasambu:
@ibacher am trying to solve this error but am somehow stackScreenshot 2021-09-30 at 7 42 41 PM. Kind help please!

cypress/support/index.ts
```diff
@@ -17,8 +17,12 @@
 import "cypress-wait-until";
 import "./commands";
 
-Cypress.on("uncaught:exception", err => {
-  console.debug("Uncaught exception", err);
+Cypress.on('uncaught:exception', (err, runnable) => {
```
ibacher, line 20: I think we should probably get rid of this "uncaught:exception" handler; it makes it much harder to see where the error actually occurs and any uncaught exceptions we should probably handle...

cypress/support/step_definitions/dictionary/addBulkConceptsToDictionary/addBulkConceptsToDictionary.ts
```diff
@@ -0,0 +1,141 @@
...
+Then('the user selects "Import existing concept"', () => {
+   cy.findByText("Import existing concept").click();
+});
+
+Then('the user should be on the "Add concepts in bulk from CIEL" page', () => {
+   cy.url().should("contain", `/users/${getUser()}/collections/${getDictionaryId()}/add/?fromSource=CIEL/`);
+});
+
+Then(
+  /the concept Id "1000" should be added to the dictionary/, conceptId => {
```
ibacher, line 74: What's the value of `conceptId` here? It would be helpful to look at the examples where we using similar arguments, maybe in other tests, of perhaps at line 79 where this works.

ibacher: To be a bit more explicit: the value of `conceptId` here is... I actually have no idea. Usually, this works because we use a regular expression with a capture group, e.g.

`/the concept Id "(1000|1001|1002)" should be in the dictionary/`

Is a regular expression that will match any of:

```
the concept Id "1000" should be in the dictionary
the concept Id "1001" should be in the dictionary
the concept Id "1002" should be in the dictionary
```

The `|` here expresses the alternating patterns and the parentheses indicate a capture group.

You can try this out in the console of your browser. For example,

```js
/my test (1000|2000|3000)/.exec("my test 1000");
```

Will give you an array like this:

```js
[ "my test 1000", "1000" ]
```

Whereas this:

```js
/my test (1000|2000|3000)/.exec("1000");
```

results in `null` (because the "my test" text is missing).

Behind the scenes, the Cypress Cucumber library will take the matching array as shown above and convert it to arguments, so that with this:

```js
Given(/the concept Id "(1000|1001|1002)" should be in the dictionary/, conceptId => {});
```

and the line in a feature file `the concept Id "1000" should be in the dictionary`, the value of the `conceptId` variable will be `1000`.

But where we don't have a regex or a capture group... I honestly don't know what the value will be. I'd expect it to be `undefined`.

That's what the failures you're pointing to about, though, is that you're taking the variable with an undefined value and passing it into a URL to construct it, which returns a 404 error because from the backend's perspective you're requesting a URL that doesn't exist.

jayasanka-sack: yes @ibacher that would be the issue.
@jwnasambu , since you are trying to pass a value to the function, we could use this:
```suggestion
  /the concept Id {string} should be added to the dictionary/, conceptId => {
```

jayasanka-sack: see:
https://github.com/TheBrainFamily/cypress-cucumber-preprocessor#reusable-step-definitions

ibacher: @jayasanka-sack Shouldn't it be `"{string}"`, i.e. we want the value to be `1000`, not `"1000"`, but if the step in the feature file is written `the concept Id "1000" should be added to the dictionary`, `{string}` should match the `"1000"` part.

jwnasambu: @ibacher and @jayasanka-sack  I have tried out the proposed changes in my terminal but still running into different errors.

jwnasambu:
@ibacher and @dkayiwa kind help please! am running into one error after another am not sure even what am doing. I kindly need help on how to solve this error please!

jayasanka-sack:
@jwnasambu I went through the implementation and the screenshots. Seems you are trying to click on a button with the text, "Add concepts", but there's no such button.
Add bulk concepts to an existing dictionary -- The user should be able to add multiple bulk concepts from CIEL (failed) (attempt 2)

cypress/support/step_definitions/dictionary/addBulkConceptsToDictionary/addBulkConceptsToDictionary.ts
```diff
@@ -0,0 +1,141 @@
...
+  cy.getConcept(`/users/${getUser()}/collections/${getDictionaryId()}/concepts/1000`).should("exist");
+});
+
+When('the user clicks the "Add concepts" floating action button', () => {
+  cy.findByTitle("Add concepts").click();
+});
+
+When('the user clicks the "Add concepts" MuiButton-label button', () => {
+  cy.findByText("Add concepts").click();
+});
```
ibacher, line 40: We definitely shouldn't have the text "MuiButton-label button" here. It should probably be `'the user clicks the "Add concepts" button'`.

cypress/support/step_definitions/dictionary/addBulkConceptsToDictionary/addBulkConceptsToDictionary.ts
```diff
@@ -0,0 +1,141 @@
...
+
+When('the user selects "Add bulk concepts"', () => {
+  cy.visit(
+    `/users/${getUser()}/collections/${getDictionaryId()}/add/?fromSource=CIEL/`
+  );
```
ibacher, line 45: Shouldn't this involve clicking on something labelled "Add bulk concepts" so we actually verify that that text works.

cypress/support/step_definitions/dictionary/addBulkConceptsToDictionary/addBulkConceptsToDictionary.ts
```diff
@@ -0,0 +1,141 @@
...
+});
+
+When(
+  /the user enters concept Id "(1000|1001|1002)"/, (conceptId) => {
+    cy.get("textarea").type(conceptId)
+  }
+); 
+
+When('the user re-enters concept Id "1000"', (conceptId) => {
```
ibacher, line 54: Why can't we just use "the user enters concept Id 1000" and reuse the above `When` step?

cypress/support/step_definitions/dictionary/addBulkConceptsToDictionary/addBulkConceptsToDictionary.ts
```diff
@@ -0,0 +1,141 @@
...
+    cy.get("textarea").type(conceptId)
+  }
+); 
+
+When('the user navigates to the "Progress notifications" page', () => {
+  cy.visit(
+    `/actions/`
+  );
+});
```
ibacher, line 63: I would suggest not navigating to the notifications to confirm anything. It's really unnecessary when we can just use `cy.getConcept(...).should("exist")` to confirm if a concept exists?

cypress/support/step_definitions/dictionary/addBulkConceptsToDictionary/addBulkConceptsToDictionary.ts
```diff
@@ -0,0 +1,141 @@
...
+
+Then('the user selects "Import existing concept"', () => {
+   cy.findByText("Import existing concept").click();
+});
```
ibacher, line 67: Any action performed by the user should be a `When` step rather than a `Then` step.

* `Given` is for how things should be when the test starts
* `When` is for what actions the user takes (the actions we are testing)
* `Then` is for the steps to see if the action did what we expected

Another way to think of it is:

* If I am saying "the user is" this should be a `Given` step, e.g. "the user is on the home page"
* If I am saying "the user does something" this should be a `When` step, e.g. "the user clicks on the red button"
* If I am saying "the user should..." this should be a `Then` step, e.g. "the user is on the new page"

cypress/support/step_definitions/dictionary/addBulkConceptsToDictionary/addBulkConceptsToDictionary.ts
```diff
@@ -0,0 +1,141 @@
...
+
+When('the user selects "Add bulk concepts"', () => {
+  cy.visit(
+    `/users/${getUser()}/collections/${getDictionaryId()}/add/?fromSource=CIEL/`
+  );
+});
+
+When(
+  /the user enters concept Id "(1000|1001|1002)"/, (conceptId) => {
+    cy.get("textarea").type(conceptId)
```
ibacher, line 50: Given how this is used, it might be necessary to make it:

```ts
.type(`${conceptId}\n`)
```

We need a separator between each concept. As it is, this types "10011002"

cypress/support/step_definitions/dictionary/addBulkConceptsToDictionary/addBulkConceptsToDictionary.ts
```diff
@@ -0,0 +1,127 @@
...
+  );
+  
+  cy.url().should(
+    "contain",
+    `/users/${getUser()}/collections/${getDictionaryId()}/add/?fromSource=CIEL/`
+  );
+});
+
+Given('CIEL concept "1000" is already in the dictionary', () => {
+  cy.getConcept(`/users/${getUser()}/collections/${getDictionaryId()}/concepts/1000`).should("exist");
```
ibacher, line 31: So these commands to get the concept need to be refactored a bit. I don't know why it didn't occur to me before now. This should actually be:

```ts
cy.getConcept(`/users/${getUser()}/collections/${getDictionaryId()}/`, "1000");
```

And similarly for the other places this is used. That will probably at least get use to the next set of errors.

cypress/support/step_definitions/dictionary/addBulkConceptsToDictionary/addBulkConceptsToDictionary.ts
jayasanka-sack: I went through the implementation and the screenshots. Seems you are trying to click on a button with the text, "Add concepts", but there's no such button.

image

ibacher (approved):

jwnasambu:
@ibacher thanks so so much for your time and the review and fix. Finally !!!

PR #730, opened 2021-08-20
OCLOMRS-1022: Answers are not being displayed even the concepts are coded

hadijahkyampeire:
Give it a backUrl which will skip the older concept version url

# JIRA TICKET NAME:
[Answers are not being displayed even the concepts are coded](https://issues.openmrs.org/browse/OCLOMRS-1022)

# Summary:
- Gave the header a backUrl which points to the dictionary concepts page to avoid going to the older concept version page which was the bug.

https://user-images.githubusercontent.com/30952856/130160565-59d6d57e-b918-4bf6-94d5-d0a14360d6a9.mov

### DIFF
--- a/src/apps/concepts/pages/ViewConceptPage.tsx
+++ b/src/apps/concepts/pages/ViewConceptPage.tsx
@@ -63,6 +63,7 @@ const ViewConceptPage: React.FC<Props> = ({
             ? concept.display_name
             : "View concept"
         }
+        backUrl={linkedDictionary ? `${linkedDictionary}concepts/` : `${dictionaryToAddTo}concepts/` }
       >
         <ProgressOverlay
           delayRender


suruchee (approved):
LGTM, thank you @hadijahkyampeire  for the work, it was a much-needed fix.

src/apps/concepts/pages/ViewConceptPage.tsx
```diff
@@ -63,6 +63,7 @@ const ViewConceptPage: React.FC<Props> = ({
             ? concept.display_name
             : "View concept"
         }
+        backUrl={linkedDictionary ? `${linkedDictionary}concepts/` : `${dictionaryToAddTo}concepts/` }
```
HerbertYiga, line 66: thanks @hadijahkyampeire  cc @dkayiwa

HerbertYiga (approved):

ibacher (approved):

ibacher:
Thanks @hadijahkyampeire! This looks like a great fix! My only question on this is are `linkedDictionary` or `dictionaryToAddTo` always going to be provided?

If both of them are undefined, it might be best to ensure the `Back` button goes back to the concept source (e.g. if I was editing a concept in the source directly rather than via a dictionary). That can be done as a separate PR, though.

hadijahkyampeire:
I agree with you @ibacher I will reproduce that scenario and ticket it up.

PR #731, opened 2021-08-20
OCLOMRS-1021: Cannot map to some sources unless logged in as a different user

hadijahkyampeire:
# JIRA TICKET NAME:
[Cannot map to some sources unless logged in as a different user](https://issues.openmrs.org/browse/OCLOMRS-1021)

# Summary:
- For where `to_source_url` is null consider `to_concept_url` and modify it to suit the source options


suruchee (commented):
Thank you @hadijahkyampeire, I can view the source while mapping answers in a concept.

PR #732, opened 2021-08-20
OCLOMRS-1028: Source disappears in answer section of a coded concept

hadijahkyampeire:
# JIRA TICKET NAME:
[Source disappears in answer section of a coded concept](https://issues.openmrs.org/browse/OCLOMRS-1028)

# Summary:
- For where `to_source_url` is null consider `to_concept_url` and modify it to suit the source options

### DIFF
--- a/src/apps/concepts/components/MappingsTableRow.tsx
+++ b/src/apps/concepts/components/MappingsTableRow.tsx
@@ -79,14 +79,7 @@ const conceptCodeFromUrl = (url: string): string => {
   return letters.join("");
 };
 
-const conceptFromUrl = (url: string): string => {
-  let letters = url.split("/");
-  letters.reverse();
-  letters.splice(0, 3, '');
-  letters.reverse();
-
-  return letters.join("/");
-};
+const conceptFromUrl = (url: string) => url.split("/").slice(0, -3).join("/") + "/";
 
 interface SourceOption extends Option {
   isInternalSource: boolean;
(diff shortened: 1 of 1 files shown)


src/apps/concepts/components/MappingsTableRow.tsx
```diff
@@ -288,6 +297,8 @@ const MappingsTableRow: React.FC<Props> = ({
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [fixedMappingType]);
 
+  const sourceUrl = toSourceUrl ? toSourceUrl : toConceptUrl ? conceptFromUrl(toConceptUrl) : undefined;
```
ibacher, line 300: Slightly less lost in ternaries:

```suggestion
  const sourceUrl = toSourceUrl ?? (!!toConceptUrl && conceptFromUrl(toConceptUrl))
```

src/apps/concepts/components/MappingsTableRow.tsx
```diff
@@ -79,6 +79,15 @@ const conceptCodeFromUrl = (url: string): string => {
...
 };
 
+const conceptFromUrl = (url: string): string => {
+  let letters = url.split("/");
+  letters.reverse();
+  letters.splice(0, 3, '');
+  letters.reverse();
+
+  return letters.join("/");
+};
```
ibacher, line 89: This could be done more efficiently using `slice()` (it also results in no array copies)

```suggestion
const conceptFromUrl = (url: string) => url.split("/").slice(0, -3).join("/") + "/";
```

ibacher (changes requested):
Thanks @hadijahkyampeire! The overall idea looks good but it seems to have broken some of the e2e tests for reasons that are a bit obscure to me. (Maybe [the screenshots](https://github.com/openmrs/openmrs-ocl-client/suites/3551444399/artifacts/85139369) are of some assistance?) I've also made some suggestions on how this could be more cleanly implemented.

hadijahkyampeire:
Thanks, @ibacher for the feedback, I have worked on it hope the tests will pass.

hadijahkyampeire:
@ibacher is there anything blocking this?

ibacher (approved):

PR #734, opened 2021-08-24
OCLOMRS-1033: Add a tab to view "My Organizations" to the organizations view

hadijahkyampeire:
# JIRA TICKET NAME:
[Add a tab to view "My Organizations" to the organizations view](https://issues.openmrs.org/browse/OCLOMRS-1033)

# Summary:
- Put back the My Organisations tab.

### DIFF
--- a/src/apps/dictionaries/constants.ts
+++ b/src/apps/dictionaries/constants.ts
@@ -2,7 +2,7 @@ import { TabType } from "../containers/types";
 
 export const TAB_LIST: TabType[] = [
   {
-    labelName: "Your Dictionaries",
+    labelName: "My Dictionaries",
     labelURL: "/user/collections/"
   },
   {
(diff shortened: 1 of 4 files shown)


suruchee (commented):
Thank you @hadijahkyampeire, for the work, LGTM. But **Your Dictionaries** and **Your Sources** have been used for dictionaries and sources so I am not sure about the **My Organizations** for organizations. What do you think, should it be consistent on all pages or not?

hadijahkyampeire:
You are right @suruchee we should look at being consistent, I think we should use `My` for dictionaries and sources, it looks like that's where we are moving. @ibacher what do you say about that?

ibacher:
@suruchee That’s a good point. Consistency is really the important thing here. 

Personally, I’ve just thought about those tabs as being “My Dictionaries” rather than “Your Dictionaries” hence how I wrote the ticket. I might lean to favouring “My” language over “Your” language, but my only real reason for that is that it’s shorter and so easier to display in the UI. 

Either “My” or “Your” are fine with me as long as we’re consistent.

ibacher (approved):
Approved with caveat that we should sort out the “My” / “Your” before merging

PR #737, opened 2021-08-31
OCLOMRS-1025:Automated test: Add and delete members in an organization

suruchee:
# JIRA TICKET NAME:
[Automated test: Add and delete members in an organization](https://issues.openmrs.org/browse/OCLOMRS-1025)
[Delete Member dialog doesn't differentiate the actual member.](https://issues.openmrs.org/browse/OCLOMRS-1057)

# Summary:
Automated test for adding and deleting members in a organization.
Fixed the delete member dialog which was deleting the wrong user.

### DIFF
--- a/cypress/support/step_definitions/organisation/deleteMember/deleteMember.ts
+++ b/cypress/support/step_definitions/organisation/deleteMember/deleteMember.ts
@@ -1,0 +1,21 @@
+import { Given, Then, When } from "cypress-cucumber-preprocessor/steps";
+import { getOrganisationId } from "../../../utils";
+
+Given("the user is on the organisation detail page", () => {
+  cy.visit(`/orgs/${getOrganisationId()}/`);
+  cy.findByText("Members").should("be.visible");
+  cy.waitUntil(() => cy.contains("li", "openmrs", { timeout: 10000 }));
+});
+When("the user clicks on the delete member button", () =>{
+  cy
+    .contains("li", "openmrs")
+    .get("li > button")
+    .first()
+    .click();
+cy.findByRole("button", { name: /Yes/i }).click();
+}
+
+);
+Then("the member should be deleted", () => {
+  cy.get("li").should("not.contain", "openmrs");
+});
--- a/cypress/support/step_definitions/organisation/deleteMember/deleteMember.ts
+++ b/cypress/support/step_definitions/organisation/deleteMember/deleteMember.ts
@@ -14,8 +14,8 @@ When("the user clicks on the delete member button", () =>{
     .click();
 cy.findByRole("button", { name: /Yes/i }).click();
 }
-
 );
+
 Then("the member should be deleted", () => {
   cy.get("li").should("not.contain", "openmrs");
 });
--- a/cypress/integration/organisation/addMember.feature
+++ b/cypress/integration/organisation/addMember.feature
@@ -1,0 +1,20 @@
+Feature: Adding an organisation Member
+  Background:
+    Given the user is logged in
+
+  @organisation
+  Scenario: The user should be able to click the button to add a new member
+    Given an organization exists
+      And the user is on the organisation detail page
+    When the user clicks on the add new member button
+    Then the user should be on the add new member dialog box
+
+  @organisation
+  @member
+  Scenario: The user should be able to add a new member
+    Given an organization exists
+      And a new user exists
+      And the user is on the add new member dialog box
+    When the user enters the member information
+      And the user submits the form
+    Then the new member should be added
--- a/cypress/integration/organisation/deleteMember.feature
+++ b/cypress/integration/organisation/deleteMember.feature
@@ -1,0 +1,14 @@
+Feature: Adding an organisation Member
+  Background:
+    Given the user is logged in
+
+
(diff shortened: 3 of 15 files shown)


cypress/support/step_definitions/organisation/deleteMember/deleteMember.ts
```diff
@@ -0,0 +1,21 @@
+import { Given, Then, When } from "cypress-cucumber-preprocessor/steps";
+import { getOrganisationId } from "../../../utils";
+
+Given("the user is on the organisation detail page", () => {
+  cy.visit(`/orgs/${getOrganisationId()}/`);
+  cy.findByText("Members").should("be.visible");
+  cy.waitUntil(() => cy.contains("li", "openmrs", { timeout: 10000 }));
+});
+When("the user clicks on the delete member button", () =>{
```
suruchee, line 9: Hi @ibacher the issue I was explaining is over here, it clicks the new member but ends up deleting the original user(the old one). Please help me to fix this.

ibacher: So this is actually a bug in the product (yay!). We use [this component](https://github.com/openmrs/openmrs-ocl-client/blob/master/src/apps/organisations/components/OrgMembers.tsx) to render the organisation members including the delete icon. Hence, [the block starting here](https://github.com/openmrs/openmrs-ocl-client/blob/c3c7d5246006bd0619ee70e947ae2d59a992d392/src/apps/organisations/components/OrgMembers.tsx#L94) where we redner each members user name and, if we are allowed to modify the organisation, the delete button.

So what happens when you click on the delete button? The [`onClick`](https://github.com/openmrs/openmrs-ocl-client/blob/c3c7d5246006bd0619ee70e947ae2d59a992d392/src/apps/organisations/components/OrgMembers.tsx#L100) handler calls `displayDeleteMemberDialog()`. `displayDeleteMemberDialog()` is actually an alias to trigger the [`showDeleteMemberDialogAction`](https://github.com/openmrs/openmrs-ocl-client/blob/c3c7d5246006bd0619ee70e947ae2d59a992d392/src/apps/organisations/components/OrgMembers.tsx#L154). Triggering that action sets the [`showDeleteMemberDialog`](https://github.com/openmrs/openmrs-ocl-client/blob/c3c7d5246006bd0619ee70e947ae2d59a992d392/src/apps/organisations/components/OrgMembers.tsx#L80) property to `true` which shows the `DeleteMemberDialog` (since `showDeleteMemberDialog` is [bound to the `open` property](https://github.com/openmrs/openmrs-ocl-client/blob/c3c7d5246006bd0619ee70e947ae2d59a992d392/src/apps/organisations/components/OrgMembers.tsx#L103)).

Simple enough and it should work. Here's the issue: there isn't just one `DeleteMemberDialog` on the page. There is one _for each member of the organisation_. So which `DeleteMemberDialog` gets show? Apparently the first one. So, no matter which delete button you click on, you are always presented with the `DeleteMemberDialog` for the first member.

Obviously, this needs to be fixed.

The obvious thing to do would be to change `showDeleteMemberDialog` into a variable that can hold a user name so that instead of having a `DeleteMemberDialog` component per member, we have just one.

E.g. something like this:

```ts
<DeleteMemberDialog
  open={showDeleteMemberDialog !== undefined}
  // NB hideDeleteMemberDialog() should now just set showDeleteMember to undefined
  handleClose={hideDeleteMemberDialog}
  handleSubmit={() => deleteMember(orgUrl, showDeleteMember)}
  user={showDeleteMember}
  orgName={orgName}
  error={deleteError}
/>
<List>
  {members?.length ? (
    members.map(m => (
      <ListItem className={classes.root} key={m.username}>
        {m.username || m.name}
        {!canModifyMembers ? null : (
          <>
            <IconButton onClick={() => displayDeleteMemberDialog(m.username)}>
              <DeleteIcon />
            </IconButton>
          </>
        )}
      </ListItem>
    ))
  ) : (
    <li>No members found!</li>
  )}
</List>
```

suruchee: Thank you so much Ian for the detail explanation and finding the bug in product. But I am little confused in the solution above so how should I define showDeleteMember as used in handelSubmit and user?

ibacher: @suruchee Oh... I meant to rename "showDeleteMemberDialog" to "showDeleteMember" (so it sounds less like a boolean value... "showDeleteMember" should be the same value as "showDeleteMemberDialog"

hadijahkyampeire: I have followed the conversation and I think we need to create a ticket to fix the modal issue, maybe we pass username instead of just a boolean, this might unblock the issue @suruchee is having. Thanks @ibacher for catching that.

hadijahkyampeire (approved):
LGTM, thanks @suruchee for the great work, and thanks @ibacher for catching the bug.

suruchee:
Thanks, @hadijahkyampeire  for the support and bug fix.

PR #738, opened 2021-08-31
OCLOMRS-1031:Your Organizations Sources shows user-owned sources

suruchee:
# JIRA TICKET NAME:
[Your Organizations Sources shows user-owned sources](https://issues.openmrs.org/browse/OCLOMRS-1031)

# Summary:
As Your Organization Source and Your Source were returning the same data, I have updated the query with verbose: true so that now it returns the desired result. And the same with dictionaries.

### DIFF
--- a/src/apps/sources/api.ts
+++ b/src/apps/sources/api.ts
@@ -22,6 +22,7 @@ const api = {
           params: {
             limit,
             page,
+            verbose: true,
             q: buildPartialSearchQuery(q),
             timestamp: new Date().getTime() // work around seemingly unhelpful caching
           }
@@ -31,6 +32,7 @@ const api = {
           params: {
             limit,
             page,
+            verbose: true,
             q: buildPartialSearchQuery(q),
             timestamp: new Date().getTime() // work around seemingly unhelpful caching
           }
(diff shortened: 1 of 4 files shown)


src/apps/sources/api.ts
```diff
@@ -22,6 +22,7 @@ const api = {
           params: {
             limit,
             page,
+            verbose: true,
```
HerbertYiga, line 25: thanks @suruchee  hope you also tested this with running OCL

hadijahkyampeire (approved):
After fixing the tests this should be merged and we test it well.

hadijahkyampeire:
@ibacher and @gracepotma please test this in staging or production to confirm that the bug was solved. @suruchee and I saw it working well through `CURL` requests for production but she could not confirm on local.

ibacher:
There may be a backend issue involved here...

So using my token when I request:

https://api.staging.openconceptlab.org/user/collections/

I get back the list of my collections

When I request:

https://api.staging.openconceptlab.org/user/orgs/collections/

I get back the list of my org collections.

However, when I request

https://api.staging.openconceptlab.org/user/orgs/collections/?q=

or 

https://api.staging.openconceptlab.org/user/orgs/collections/?customValidationSchema=OpenMRS

With or without `verbose=true` I still get dictionaries that belong to me rather than to my organisations.

PR #740, opened 2021-09-07
OCLOMRS-1015: Only add concepts mapped as answers or concept set members

hadijahkyampeire:
# JIRA TICKET NAME:
[ Only add concepts mapped as answers of concept set members](https://issues.openmrs.org/browse/OCLOMRS-1015)

# Summary:
- Only add concepts mapped as answers or concept sets.

### DIFF
--- a/src/apps/dictionaries/logic.ts
+++ b/src/apps/dictionaries/logic.ts
@@ -13,6 +13,13 @@ const fetchConceptMappings = async (
   }
 };
 
+const filterUnneededMappings = (mappings: APIMapping[]) =>
+  mappings
+    .filter(m => m.to_concept_url !== undefined)
+    .filter(
+      m => m.map_type === "Q-AND-A" || m.map_type === "CONCEPT-SET"
+    ) as InternalAPIMapping[];
+
 const recursivelyFetchToConcepts = async (
   fromSource: string,
   fromConceptIds: string[],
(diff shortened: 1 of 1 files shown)


src/apps/dictionaries/logic.ts
```diff
@@ -50,6 +50,7 @@ const recursivelyFetchToConcepts = async (
   const loadedConcepts = new Set();
   for (let i = 0; i < levelsToCheck; i += 1) {
     const toConceptCodes = mappingsLists[i]
+      .filter(m => m.map_type === 'Q-AND-A' || m.map_type === 'CONCEPT-SET')
```
HerbertYiga, line 53: thanks @hadijahkyampeire  cc @ibacher

ibacher:
@hadijahkyampeire Did you test this code change? Here's what I did:

1. Create a new source
2. In the new source add two concepts TRUE and YES
3. Map TRUE as "Same as" YES
4. Create a new dictionary
5. Add TRUE to the dictionary
6. Note that YES is also added to the dictionary

This appears to indicate that it's not doing what was required by the ticket. There may be more steps that need to be covered than my initial analysis.

ibacher:
PS I'm wondering if maybe what's left to do has to do with the fact that we send something `cascase=sourcemappings` when adding the reference to the collection [API docs](https://docs.openconceptlab.org/en/latest/oclapi/apireference/collections.html#add-a-reference-to-a-collection).

hadijahkyampeire:
Thanks @ibacher for the review honestly speaking, I admit that I am confused about the ticket requirement, I wanted to request for a call but I know you are still busy with the cloning work, but maybe next week if you get time please give me a few minutes and we chat about this. There are also some automated tests that are beating my understanding maybe we can brainstorm about how best I can approach them.

ibacher:
@hadijahkyampeire Yes, let's set up a call for early next week to talk through that.

ibacher (approved):

PR #743, opened 2021-09-14
OCLOMRS-1047:List of concepts - Text search and filters don't work anymore

suruchee:
# JIRA TICKET NAME:
[Clear Filters button does not work well](https://issues.openmrs.org/browse/OCLOMRS-1047)

# Summary:
- Fixes the Clear Filter button functionality

### DIFF
--- a/src/apps/concepts/components/FilterOptions.tsx
+++ b/src/apps/concepts/components/FilterOptions.tsx
@@ -17,6 +17,7 @@ import {
 } from "@material-ui/core";
 import { CONCEPT_CLASSES, DATA_TYPES, CONCEPT_GENERAL } from "../../../utils";
 import { Link } from "react-router-dom";
+import { useHistory } from "react-router";
 
 const useStyles = makeStyles((theme: Theme) =>
   createStyles({
@@ -163,12 +164,14 @@ const FilterOptions: React.FC<FilterOptionsProps> = ({
 }) => {
   const classes = useStyles();
   const [searchText, setSearchText] = useState("");
+  const { replace: goTo } = useHistory();
 
   const clearAllFilters = () => {
     setCheckedClasses([]);
     setCheckedDataTypes([]);
     setCheckedGeneral([]);
     setCheckedSources([]);
+    goTo(url);
   };
 
   return (
--- a/src/apps/concepts/pages/ViewConceptsPage.tsx
+++ b/src/apps/concepts/pages/ViewConceptsPage.tsx
@@ -139,7 +139,11 @@ const ViewConceptsPage: React.FC<Props> = ({
 
   const { replace: goTo } = useHistory(); // replace because we want to keep the back button useful
   const { pathname: url } = useLocation();
-  const containerUrl = url.replace("/concepts", "");
+  const [containerUrl, setContainerUrl] = useState("");
+
+  useEffect(() => {
+    setContainerUrl(url.replace("/concepts", ""));
+  }, [url]);
 
   const { ownerType, owner } = useParams<{
     ownerType: string;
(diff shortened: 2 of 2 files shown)


src/apps/concepts/pages/ViewConceptsPage.tsx
```diff
@@ -239,37 +235,50 @@ const ViewConceptsPage: React.FC<Props> = ({
...
-        sortDirection: sortDirection,
-        sortBy: sortBy,
-        dataTypeFilters: initialDataTypeFilters,
-        classFilters: initialClassFilters,
-        sourceFilters: initialSourceFilters,
-        includeRetired: initialGeneralFilters.includes("Include Retired"),
-        includeAdded: generalFilters.includes("Include Added Concepts")
-      });
-    }
+    retrieveConcepts({
```
HerbertYiga, line 245: thanks @suruchee  have you also tested on a working instance of ocl

hadijahkyampeire (commented):
Thanks for the fix @suruchee , it is working well but I tried to keep most of what we had and only add the necessary dependencies to the `useEffect` and it still works, let's do that so that we don't break some other parts.

src/apps/concepts/pages/ViewConceptsPage.tsx
```diff
@@ -239,37 +235,50 @@ const ViewConceptsPage: React.FC<Props> = ({
...
+      includeRetired: initialGeneralFilters.includes("Include Retired"),
+      includeAdded: generalFilters.includes("Include Added Concepts")
+    });
     // i don't know how the comparison algorithm works, but for these arrays, it fails.
     // stringify the arrays to work around that
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [retrieveConcepts, retrieveDictionary, retrieveSource, containerUrl]);
-
+  }, [
+    retrieveConcepts,
```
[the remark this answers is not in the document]
suruchee: Thanks, @hadijahkyampeire, I have updated with your suggestions.

hadijahkyampeire:
Also, the `ClearFilters` button is not working, we should create a follow-up ticket to have that work again cc @suruchee

hadijahkyampeire (approved):
LGTM, thanks for the fix.

hadijahkyampeire:
@suruchee this same work has been covered in @jwnasambu [PR](https://github.com/openmrs/openmrs-ocl-client/commit/04efed8b71301a37430cdceabe6176432c80a2d1), you should rebase and maybe make this PR fixed the clear filters bug.

suruchee:
Hi @ibacher  and @hadijahkyampeire, the clear all button works with the code changes but only works in double click, I also tried changing dependency in viewConceptsPage but did not work. Can you guys have a look?

src/apps/concepts/components/FilterOptions.tsx
```diff
@@ -163,12 +164,14 @@ const FilterOptions: React.FC<FilterOptionsProps> = ({
...
   const classes = useStyles();
   const [searchText, setSearchText] = useState("");
+  const { replace: goTo } = useHistory();
 
   const clearAllFilters = () => {
     setCheckedClasses([]);
     setCheckedDataTypes([]);
     setCheckedGeneral([]);
     setCheckedSources([]);
+    goTo(url);
```
ibacher, line 174: I actually wonder about this... Right now the behaviour is that I click each filter and then click the "APPLY FILTERS" link to actually update the view. The logic behind the way the "CLEAR ALL" button currently work is that it's just clearing the filters as if you had cleared them manually. You still need to click "APPLY FILTERS" for the changes to make.

I think the question is: is the user experience expectation that clicking on the "CLEAR ALL" button removes all the filters *and* updates the table, or do we want to just clear all the filters and let the updating of the table take place after the user has reset their filters.

There are two reasons I might click on the "CLEAR ALL" button:

1. I really want to go back to the view of all concepts.
2. I just want to clear out the filters I previously had so I can create a new set of filters.

This new behaviour obviously makes the flow easy for 1, but is it the right flow for 2?

And in either case, I would update this PR to reflect what this change actually does (i.e. reloads the table when clicking "CLEAR ALL") rather than the now disconnected ticket it was first attached to.

hadijahkyampeire: I think the "CLEAR Filter" button should remove all the filters, update the url and make a request since the URL has changed so that the user goes back to the view without filters but we should remember to keep `page and q` since they are not attached to the other filters.

ibacher: `q` makes sense to keep, but not `page` if we're making a new query (why would I want to be on page 5 of all the concepts after having been on page 5 of all the question concepts?)

hadijahkyampeire: That would be great @ibacher if you can update this PR with that fix, I will update the ticket description to capture the clear filter changes.

hadijahkyampeire: You are right @ibacher , now I see it for the `page` setting to `1`, though given where the clear filters button is placed it looks like it's only for clearing the filters in the sideBar and not anything else. But I think that's a minor UX thing, now that I think of it, it's probably what the user expects.

suruchee: Thanks, @hadijahkyampeire, and @ibacher I think we should leave it as it is for now. So, just closing the PR.

PR #747, opened 2021-09-23
OCLOMRS-991: Edit and retire a concept

hadijahkyampeire:
# JIRA TICKET NAME:
[The Basic User: Editing and retiring a concept.](https://issues.openmrs.org/browse/OCLOMRS-991)

# Summary:
- Edit and retire a concept

### DIFF
--- a/cypress/integration/concept/edit.feature
+++ b/cypress/integration/concept/edit.feature
@@ -1,0 +1,25 @@
+Feature: Edit and retire a custom concept
+        Background:
+            Given the user is logged in
+
+        @dictionary
+        @concept
+        @organisation
+        Scenario: The user should be able edit a custom concept
+            Given a dictionary exists
+              And a concept exists
+              And the user is on the edit concept page
+             When the user edits the concept name
+            #   And the user submits the form
+            #  Then the concept should be updated
+
+        # @dictionary
+        # @concept
+        # @organisation
+        # Scenario: The user should be able to retire a custom concept
+        #     Given a dictionary exists
+        #       And a concept exists
+        #       And the user is on the edit concept page
+        #      When the user clicks the Menu button
+        #       And the user selects the "Retire concept" menu list item
+        #      Then the concept should be retired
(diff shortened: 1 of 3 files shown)


hadijahkyampeire:
@ibacher I am having this issue with the concept not being edited, it worked the first time it ran but after that, it started failing, I am not sure why that is happening, please take a look and point me in the right direction. Thanks

ibacher:
@hadijahkyampeire Sorry for the late response. I appear to have slightly lost track of this one.

The problem is that the test to retire the concept succeeds while the test to edit the concept fails. The simple reason for this is just that we require more fields than our supplied in your test concept, which isn't a problem when retiring it (there's no real need for a validation check on a retired concept), but is a problem when editing it.

It's likely that this means we need to make some adjustments to what the `createConcept` command takes as arguments, since it was really designed to load a very minimal concept.

In particular, I think we need an `external_id` (which can be set automatically) and a datatype. Note that you don't need to explicitly pass `getConceptId()` into the command either... that's the default.

hadijahkyampeire:
> @hadijahkyampeire Sorry for the late response. I appear to have slightly lost track of this one.
> 
> The problem is that the test to retire the concept succeeds while the test to edit the concept fails. The simple reason for this is just that we require more fields than our supplied in your test concept, which isn't a problem when retiring it (there's no real need for a validation check on a retired concept), but is a problem when editing it.
> 
> It's likely that this means we need to make some adjustments to what the `createConcept` command takes as arguments, since it was really designed to load a very minimal concept.
> 
> In particular, I think we need an `external_id` (which can be set automatically) and a datatype. Note that you don't need to explicitly pass `getConceptId()` into the command either... that's the default.

Thanks @ibacher I will add `externale_id` and `datatype` to the create command and see how that goes.

hadijahkyampeire:
@ibacher I have added all the required fields but it still behaves the same, it passes on the first run and fails. Also, I wonder why it doesn't throw an error.

ibacher:
@hadijahkyampeire Alright... yeah, I'm seeing what you're seeing. I really have no explanation. Neither when it succeeds nor when it fails does it actually submit the form or update the concept at all.

hadijahkyampeire:
Maybe I need to check for the updated field and see if its value has changed.

PR #749, opened 2021-09-24
OCLOMRS-1054:Avoid implying SYNONYM is a name type

suruchee:
# JIRA TICKET NAME:
[Avoid implying SYNONYM is a name type](https://issues.openmrs.org/browse/OCLOMRS-1054)

# Summary:
Name type of Synonym is displayed at the top in grey italic text with a bracket

### DIFF
--- a/src/apps/concepts/components/NamesTable.tsx
+++ b/src/apps/concepts/components/NamesTable.tsx
@@ -47,8 +47,7 @@ const useStyles = makeStyles((theme: Theme) =>
       minWidth: "150px"
     },
     menuItemSynonym: {
-      color: "#A9A9A9",
-      fontStyle: "italic"
+      opacity: 0.38
     }
   })
 );
@@ -148,13 +147,15 @@ const NamesTable: React.FC<Props> = ({
                       >
                         {NAME_TYPES.map(nameType => (
                           <MenuItem key={nameType.value} value={nameType.value}>
-                            {nameType.label.includes("Synonym") ? (
-                              <div className={classes.menuItemSynonym}>
-                                ({nameType.label})
-                              </div>
-                            ) : (
-                              <div>{nameType.label}</div>
-                            )}
+                            <div
+                              className={
+                                nameType.label === "Synonym"
+                                  ? classes.menuItemSynonym
+                                  : undefined
+                              }
+                            >
+                              ({nameType.label})
+                            </div>
                           </MenuItem>
                         ))}
                       </Field>
--- a/src/apps/concepts/components/NamesTable.tsx
+++ b/src/apps/concepts/components/NamesTable.tsx
@@ -45,6 +45,10 @@ const useStyles = makeStyles((theme: Theme) =>
     },
     minimumCellWidth: {
       minWidth: "150px"
+    },
+    menuItemSynonym: {
+      color: "#A9A9A9",
+      fontStyle: "italic"
     }
   })
 );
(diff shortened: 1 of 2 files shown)


src/apps/concepts/components/NamesTable.tsx
```diff
@@ -144,7 +148,13 @@ const NamesTable: React.FC<Props> = ({
...
                         {NAME_TYPES.map(nameType => (
                           <MenuItem key={nameType.value} value={nameType.value}>
-                            {nameType.label}
+                            {nameType.label.includes("Synonym") ? (
+                              <div className={classes.menuItemSynonym}>
+                                ({nameType.label})
+                              </div>
+                            ) : (
+                              <div>{nameType.label}</div>
+                            )}
```
ibacher, line 157: Two simplifications:

```suggestion
                            <div className={nameType.label === "Synonym" && classes.menuItemSynonym}>
                              ({nameType.label})
                            </div>
```

1. We only want this to apply to "Synonym" not "Custom Synonym" (since presumbably that needs some kind of value).
2. We can do this with a single `div`

src/utils/constants.ts
```diff
@@ -268,8 +268,8 @@ export const DATA_TYPES: string[] = [
 ];
 
 export const NAME_TYPES: { [key: string]: string }[] = [
-  { value: "FULLY_SPECIFIED", label: "Fully Specified" },
   { value: "null", label: "Synonym" }, // value here should be null but inputs html can't handle null as a value
+  { value: "FULLY_SPECIFIED", label: "Fully Specified" },
```
ibacher, line 272: I'm less sure about this... "Fully Specified" should definitely be the first option we display when creating a new name, so maybe we just don't change this for now?

src/apps/concepts/components/NamesTable.tsx
```diff
@@ -45,6 +45,10 @@ const useStyles = makeStyles((theme: Theme) =>
     },
     minimumCellWidth: {
       minWidth: "150px"
+    },
+    menuItemSynonym: {
+      color: "#A9A9A9",
+      fontStyle: "italic"
```
ibacher, line 51: I'm not sure about the font style. Also, since we use the Material Design System for this app, we should consult their guidelines about color. In particular, for disabled text, instead of adjusting the color directly, the [Material guidelines](https://material.io/design/color/text-legibility.html#text-backgrounds) suggest using 38% percent opacity.

That means this should probably be:

```suggestion
      opacity: 0.38
```

ibacher (changes requested):

src/utils/constants.ts
ibacher: Incidentally, I think this is why there's a test failure. The test assumes that the first name entered is always a FSN and for any concept in any language we always require at least one FSN

src/apps/concepts/components/NamesTable.tsx
suruchee: Hi @ibacher, I tried this with a single div but ended up with an error that says: Type 'string | false' is not assignable to type 'string | undefined'.   Type 'false' is not assignable to type 'string | undefined'.

src/utils/constants.ts
suruchee: Thanks, for catching this test failure, I will revert it.

suruchee:
@ibacher, I have updated with your suggestions please have a look.

src/apps/concepts/components/NamesTable.tsx
ibacher: Try this expression instead:

```ts
nameType.label === "Synonym" ? classes.menuItemSynonym : undefined
```

suruchee: @ibacher , it's updated please have a look.

ibacher (approved):
LGTM! Thanks @suruchee!

PR #750, opened 2021-09-28
OCLOMRS-1056:Unified OCL and OpenMRS Dictionary Manager Experience

suruchee:
# JIRA TICKET NAME:
[Unified OCL and OpenMRS Dictionary Manager Experience](https://issues.openmrs.org/browse/OCLOMRS-1056)

# Summary:
AppIcon is added in the header and three buttons to link to OCL TermBrowser,  OpenMRS Dictionary Manager, and Bulk importer.


src/components/AppsMenu.tsx
```diff
@@ -0,0 +1,139 @@
...
+                      <Box className={classes.box}>
+                        <OpenMRSLogo />
+                        <Typography className={classes.text}>
+                          OpenMRS <br /> Dictionary <br /> Manager
+                        </Typography>
+                      </Box>
+                    </Link>
+                  </MenuItem>
+                  <MenuItem onClick={handleClose} className={classes.menuItem}>
+                    <a
```
hadijahkyampeire, line 119: Can we make these <a> tags open in a new tab so that the user does not completely loose their previous page? Just adding ```target='_blank' rel='noreferrer'``` will do that.

src/components/AppsMenu.tsx
```diff
@@ -0,0 +1,139 @@
...
+                      <Box className={classes.box}>
+                        <Web fontSize="large" />
+                        <Typography className={classes.text}>
+                          OCL <br /> TermBrowser
+                        </Typography>
+                      </Box>
+                    </a>
+                  </MenuItem>
+                  <MenuItem onClick={handleClose} className={classes.menuItem}>
+                    <Link to="/" className={classes.link}>
```
hadijahkyampeire, line 109: I am not sure why we need to link to exactly where the user is because this is what this link does, just wondering.

src/components/AppsMenu.tsx
```diff
@@ -0,0 +1,139 @@
...
+            {...TransitionProps}
+            style={{
+              transformOrigin: "right top"
+            }}
+          >
+            <Paper>
+              <ClickAwayListener onClickAway={handleClose}>
+                <MenuList autoFocusItem={open} id="menu-list-grow">
+                  <MenuItem onClick={handleClose} className={classes.menuItem}>
+                    <a href={getOCLURL()} className={classes.link}>
```
hadijahkyampeire, line 99: Also make this open in a new tab.

hadijahkyampeire (commented):
Nice work @suruchee , just a few suggestions, and UI/UX improvements. Also, I looked at the document with the mocks, we are missing some styles in this PR, I will attach some screenshots to show what I think we should have.

hadijahkyampeire:
This is what I see on this PR.
Screenshot 2021-09-28 at 14 01 36
This is what I think we need to make it so that it looks much more like what is in the mocks.
Screenshot 2021-09-28 at 13 54 38

Just adding some background color, changing the color of the icon, and increasing on the margin-top for the menu box. Let me know if you need any help with any of these.

src/components/AppsMenu.tsx
suruchee: Thanks for your review @hadijahkyampeire, I have updated as per your suggestions. I missed a few of the UI changes by looking into the OCL site.

src/components/AppsMenu.tsx
suruchee: This is to create a consistent UI feel for users between OCL and dictionary manager as you can view it in https://app.openconceptlab.org/#/ apps menu

hadijahkyampeire (approved):
Thanks for the changes @suruchee , LGTM

hadijahkyampeire:
@ibacher how do you see this?

src/components/AppsMenu.tsx
ibacher: It's probably preferable to use the MaterialUI `Link` class here.

src/components/OpenMRSLogo.tsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+
+const OpenMRSLogo: React.FC<React.DetailedHTMLProps<
+  React.HTMLAttributes<HTMLDivElement>,
+  HTMLDivElement
+  >> = () => {
+  return (
+      <img src="/openmrs_logo_white.gif"  alt={"openmrs"} style={{width:"30px"}} />
```
ibacher, line 8: Probably best to use a SVG rather than the GIF here. @suruchee: I've sent you an SVG of the OMRS logo via Slack.

src/utils/utils.ts
```diff
@@ -119,3 +119,24 @@ export function stableSort<T>(array: T[], comparator: (a: T, b: T) => number) {
   });
   return stabilizedThis.map(el => el[0]);
 }
+// @ts-ignore
+export const getEnv = forURL => {
```
ibacher, line 123: 1. I don't like anything where we just use a blanket `//@ts-ignore`.
2. Rather than this, it's probably better to use the `OCL_SIGNUP_URL` constant we already have or even just evolve that to take the OCL application server name and we can generate the signup URL and this part from it.

ibacher (changes requested):
Thanks @suruchee for this! I've made some small comments on a couple of things. However, I have some slightly larger comments focused on where we're placing this and how this looks:

1. Why don't we just put this in the left nav bar the way the OCL app does?
2. The icons should be vertically stacked rather than horizontally spread, as this leads to a better mobile experience (and occupies less screen real estate).

suruchee:
> Thanks @suruchee for this! I've made some small comments on a couple of things. However, I have some slightly larger comments focused on where we're placing this and how this looks:
> 
> 1. Why don't we just put this in the left nav bar the way the OCL app does?
> 2. The icons should be vertically stacked rather than horizontally spread, as this leads to a better mobile experience (and occupies less screen real estate).

@ibacher, I could not understand your comment about the UI, as the current OCL UI is  
Screenshot from 2021-09-29 10-47-11

ibacher:
@suruchee I was talking about this:

OCL Tools Menu

(It's what you get if you click on the wrench while the toolbar is collapsed)

suruchee:
@ibacher, do you mean to remove the appIcon and only have the wrench in the left navbar as it is in OCL or to include both?

suruchee:
@ibacher, do you mean to remove the appIcon and only have the wrench in the left navbar as it is in OCL or to include both?

ibacher:
@suruchee I'd be in favour of removing the app icon. This is, I think, more in keeping with [Material Design's](https://material.io/design/layout/understanding-layout.html#layout-anatomy) overall layout that tries to keep navigation all in the left bar and reserve the spots on the app bar for [in-app actions](https://material.io/components/app-bars-top#anatomy).

ibacher:
PS Though on this that I hadn't expressed: we should still use that grid icon to indicate visually that it is an app switcher and not the wrench icon.

suruchee:
@ibacher, please review the updated change.

src/utils/constants.ts
```diff
@@ -1,5 +1,7 @@
 import { Option } from "./types";
 
+export const OCL_URL = "https://app.openconceptlab.org/";
```
ibacher, line 3: So, my suggestion here was meant to work more like the constants below, that is:

```ts
export const OCL_URL: string =
  // @ts-ignore OCL_URL is injected at runtime
  window.OCL_URL ||
  "https://app.staging.openconceptlab.org/#/";
```

Then we can adjust `OCL_SIGNUP_URL` to be defined this way:

```ts
export const OCL_SIGNUP_URL = OCL_URL + "accounts/signup";
```

The properties defined here are injected via environment variables so that we can drive locations by configuration rather than by hard-coding like `getEnv()` does.

ibacher (changes requested):
@suruchee Nice work! I've got a few more things, largely to ensure that we avoid as much hard-coding of both layouts and URLs as we can.

suruchee:
@ibacher, I have changed the position of the menu but it worked without absolute position but had to add  **overflowY: "unset"** in navOpen to avoid scroll, please have a look.

ibacher (approved):
Thanks Suruchi! One very small thing and I think this looks great!

PR #751, opened 2021-10-17
OCLOMRS-1069: Scroll position issue when navigating between dictionary results

merovingienne:
# JIRA TICKET NAME:
[OCLOMRS-1069: Scroll position issue when navigating between dictionary results](https://issues.openmrs.org/browse/OCLOMRS-1069)

# Summary:
When navigating between pages of dictionary results, the page remains at the last scrolled position. It should move to top when going to a different page of dictionaries.

### DIFF
--- a/src/apps/dictionaries/components/ViewDictionariesPage.tsx
+++ b/src/apps/dictionaries/components/ViewDictionariesPage.tsx
@@ -1,20 +1,19 @@
 import React, { useEffect } from "react";
-import { APIDictionary } from "../types";
-import { ProgressOverlay } from "../../../utils/components";
-import { useQueryParams } from "../../../utils";
 import { useHistory, useLocation } from "react-router";
-import qs from "qs";
-import ViewDictionaries from "../components/ViewDictionaries";
-import {
-  retrievePublicDictionariesAction,
-  toggleShowVerifiedAction
-} from "../redux";
+import { Link } from "react-router-dom";
 import { Fab, Tooltip } from "@material-ui/core";
 import { Add as AddIcon } from "@material-ui/icons";
-import { Link } from "react-router-dom";
-import { ContainerOwnerTabs } from "../../containers/components";
+
+import ViewDictionaries from "../components/ViewDictionaries";
 import { Header } from "../../../components";
+import { ContainerOwnerTabs } from "../../containers/components";
+import { retrievePublicDictionariesAction, toggleShowVerifiedAction } from "../redux";
+
 import { TAB_LIST, PER_PAGE, TITLE } from "../constants";
+import { useQueryParams, generateURLWithQueryParams } from "../../../utils";
+import { ProgressOverlay } from "../../../utils/components";
+
+import { APIDictionary } from "../types";
 
 interface Props {
   loading: boolean;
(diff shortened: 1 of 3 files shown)


src/utils/utils.ts
```diff
@@ -1,6 +1,8 @@
 import { LOCALES } from "./constants";
 import { snakeCase } from "lodash";
 
+export * from "./urlUtils";
+
```
ibacher, line 5: Why here and not in `src/utils/index.ts`?

src/apps/dictionaries/components/ViewDictionariesPage.tsx
```diff
@@ -53,13 +52,15 @@ const ViewPublicDictionariesPage: React.FC<Props> = ({
...
 
-  const gimmeAUrl = (params: { page?: number; q?: string }) => {
-    const newParams: { page?: number; q?: string } = {
-      ...queryParams,
-      ...params
-    };
-    return `${url}?${qs.stringify(newParams)}`;
-  };
+  const changePage = (params: { page?: number; q?: string }) => {
+    const targetURL = generateURLWithQueryParams(url, params);
```
ibacher, line 56:
```suggestion
    const targetURL = generateURLWithQueryParams(url, params, queryParams);
```

src/apps/dictionaries/components/ViewDictionariesPage.tsx
```diff
@@ -1,20 +1,19 @@
...
+import ViewDictionaries from "../components/ViewDictionaries";
 import { Header } from "../../../components";
+import { ContainerOwnerTabs } from "../../containers/components";
+import { retrievePublicDictionariesAction, toggleShowVerifiedAction } from "../redux";
+
 import { TAB_LIST, PER_PAGE, TITLE } from "../constants";
+import { useQueryParams, generateURLWithQueryParams } from "../../../utils";
+import { ProgressOverlay } from "../../../utils/components";
+
+import { APIDictionary } from "../types";
```
ibacher, line 16: It's not really a good idea to do this kind of refactoring in a ticket that isn't about refactoring, especially as it potentially breaks other PRs and tends to obscure what this actually does

merovingienne: @ibacher Understood. Shall we create a separate code refactor epic in that case? We can improve the developer experience by introducing some order to the imports. Some suggestions:

* Separate imports of third-party libraries and our files with a single empty line (maybe when there are more than 3 total imports?)
* Sort third-party imports by order of importance
(eg: core imports like `react`, `react-router-dom` first, `material-ui` and `redux` next, and util functions from `lodash` next)
* Sort named imports within curly braces alphabetically.
eg: `import { applyTransformation, calculateTotals, getFormulation } from "some-source";`

src/utils/utils.ts
merovingienne: I actually missed that, my bad. Will put this line in `src/utils/index.ts`.

Side-note: the filename of `src/utils/utils.ts` is not ideal. It's currently a collection of miscellaneous functions. I don't have a good alternative in mind right now, though.

src/apps/dictionaries/components/ViewDictionariesPage.tsx
hadijahkyampeire: I agree with you @merovingienne, I use that in some other project I work on and it is indeed a good coding convention.

hadijahkyampeire (approved):
This LGTM, thanks @merovingienne

src/apps/dictionaries/components/ViewDictionariesPage.tsx
ibacher: @merovingienne 100% approve of adding an epic to do that. I might suggest using something like [this prettier plugin](https://github.com/trivago/prettier-plugin-sort-imports) so we can automate this across files.

PR #754, opened 2021-10-19
OCLOMRS-1053: Concept number mismatch between concept list and dictionary view

merovingienne:
# JIRA TICKET NAME:
[OCLOMRS-1053: Concept number mismatch between concept list and dictionary view](https://issues.openmrs.org/browse/OCLOMRS-1053)

# Summary:
The number of total concepts shown in the dictionary view page does not match with the number in the concept list page.


merovingienne:
@ibacher kindly take a look when possible.

src/apps/dictionaries/api.ts
```diff
@@ -11,6 +11,10 @@ import { default as containerAPI } from "../containers/api";
 
 const api = {
   ...containerAPI,
+  conceptCounts: {
+    retrieve: (dictionaryUrl: string, source?: string) =>
+      unAuthenticatedInstance.head(`${dictionaryUrl}concepts`, { params: { source } }),
```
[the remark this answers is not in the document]
merovingienne: Ah I see. I'll wait until we resolve the other discussion on createAsyncThunk task response type before integrating this.

merovingienne:
@ibacher This has been a veritable rabbit hole. I found a couple of issues.

**Tl;dr - Fixing the first problem caused some existing tests to fail.**

1. In the `AuthenticationRequired` component, prop changes caused entire child component subtrees to mount, unmount and remount. This caused multiple API calls. Changed it so that children don't mount until profile data are loaded.

2. Concept counts API call URL lacked a trailing slash. This caused the API server to redirect the request, increasing call time.

Once these two were fixed, it seems some tests that unknowingly depended on the remount behaviour are now broken.

For future reference (mostly for myself):

Previously, loading profile data caused component remounts. This allowed Formik forms down the hierarchy to remount, initialising them with proper values from API calls. But without the remounts, the forms were initialised with empty values while fetching relevant data. This caused errors as Formik doesn't reinitialise by default, when the initial values variable changes.

eg: Edit Org privacy test

Fixed this by delaying Org Edit form render until org details are fetched.

Have a couple more tests to fix. Will inform once done.

sherrif10:
cc @merovingienne , Is this a blocker?

merovingienne:
@sherrif10 I'm debugging the tests currently. If I'm not mistaken, in the Add Concepts to Dictionary test group, the last 3 cases (tests 2-4) cannot succeed as they require a dictionary to exist first, but we are not enforcing that requirement in the tests. @ibacher could you kindly confirm this?

I'm testing now with a dictionary first created before the body of these tests. There's an API call failure as well (retrieveConcepts API call - URL excluding added concepts throws an HTTP 500).

Any help/tips would be appreciated!

merovingienne:
Update: The retrieveConcepts API call succeeds when pointing to the OCL QA server. This means the dockerised server used for testing has a bug. I just saw that there's an updated image. Will test with that and update.

cypress/integration/dictionary/addConceptsToDictionary.feature
```diff
@@ -14,7 +14,8 @@ Feature: Add concepts to an existing dictionary
   @dictionary
   @ciel
   Scenario: The user should be able to add a single concept from their preferred source
-    Given the user is on the "Import existing concept" page
+    Given a dictionary exists
```
sherrif10, line 17: is there any good way how we can refactor these scenarios, Especially scenarios which look the same.

ibacher: Yeah I think this "Given a dictionary exist" step belongs in the backgroun of these tests

ibacher:
> but we are not enforcing that requirement in the tests. @ibacher could you kindly confirm this?

Yes, that seems to be the case. Sorry for the difficulty!

cypress/integration/dictionary/addConceptsToDictionary.feature
ibacher: I actually think this is fine. In general, I'm happy with the feature files being a little repetitive, as each feature should uniquely describe the steps it takes to use that feature, even if they are obvious or done everywhere.

merovingienne:
I've fixed the integration tests now. But one unit test is giving me trouble. Essentially, the failing unit test needs to wait before the page loads according to the defined API call sequence (component load state changes).

@ibacher I think it's time I seek your help. Could you kindly take a look at it?

merovingienne:
@ibacher I've fixed the unit test failure by not delaying the content render in the View Concepts page. This (hopefully) allows us to close this ticket.

I'd feel better if you could read through the logic in that file and see if it's comprehensible. I'm worried if the component state machine is confusing.
https://github.com/openmrs/openmrs-ocl-client/pull/754/files#diff-60efd061211b613f4964e3afe828937492f8764f1f49790b47eb8cb19a778bbb

merovingienne:
@ibacher @suruchee  kind reminder on this.

gracepotma:
Ping @suruchee, any thoughts/concerns on this?

PR #755, opened 2021-10-20
OCLOMRS-1051:Pick concepts - Choice issue with same IDs

suruchee:
# JIRA TICKET NAME:
[Pick concepts - Choice issue with same IDs](https://issues.openmrs.org/browse/OCLOMRS-1051)

# Summary:
Issue fixed for choosing concepts with same IDs while picking concepts

### DIFF
--- a/src/apps/concepts/components/ConceptsTableRow.tsx
+++ b/src/apps/concepts/components/ConceptsTableRow.tsx
@@ -78,7 +78,7 @@ const actionIcon = (
 };
 
 const conceptNameCell = (
-  toggleSelect: (event: React.MouseEvent<unknown>, id: string) => void,
+  toggleSelect: (event: React.MouseEvent<unknown>, uuid: string) => void,
   row: APIConcept,
   linkedDictionary?: string,
   dictionaryToAddTo?: string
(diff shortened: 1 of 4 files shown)


hadijahkyampeire (approved):
LGTM, thanks @suruchee

src/apps/concepts/components/ConceptsTableRow.tsx
```diff
@@ -78,16 +78,16 @@ const actionIcon = (
...
   linkedDictionary?: string,
   dictionaryToAddTo?: string
 ) => {
   let link = row.version_url;
   const params = new URLSearchParams();
   if (linkedDictionary) {
-    params.append("linkedDictionary", linkedDictionary);
-  }
+                          params.append("linkedDictionary", linkedDictionary);
+                        }
```
ibacher, line 90:
```suggestion
    params.append("linkedDictionary", linkedDictionary);
  }
```

ibacher (commented):
I think that instead of having `concept.uuid || ""` all over the place, we can just make `uuid` a required field on APIConcept without any issues. APIConcept (as the name suggests) is supposed to represent a concept as retrieved from the API and those should always have uuids (because the uuid field is actually the database identifier).

ibacher (approved):

PR #760, opened 2021-10-26
OCLOMRS-1065: Upgrade the MUI to version 5 in dictionary manager

suruchee:
# JIRA TICKET NAME:
[Upgrade the mui to version 5 in dictionary manager](https://issues.openmrs.org/browse/OCLOMRS-1065)

# Summary:
Upgrade to MaterialUI 5


package.json
```diff
@@ -12,9 +12,12 @@
     "npm": ">=6.4.1"
   },
   "dependencies": {
-    "@material-ui/core": "^4.11.3",
-    "@material-ui/icons": "^4.11.2",
-    "@material-ui/lab": "^4.0.0-alpha.57",
+    "@emotion/react": "^11.4.1",
+    "@emotion/styled": "^11.3.0",
```
ibacher, line 16: Do we actually need "@emotion/react" and "@emtion/styled"? I know that MUI is using them under the hood, but we don't seem to be actually using them anywhere and it's better to keep the dep list as minimal as possible.

ibacher (approved):
Aside from the minor comment I've left, I'm happy for this to be merged in.

package.json
suruchee: Thank you so much @ibacher, for the review. I had not used emotion at the beginning but it ended up with some error **Unhandled Rejection error** in Reducer. And was fixed by adding emotion so had to add this dep.

ibacher:
Where we have to use test ids, can we try using `findByLabelText()`. The new version of MUI moves the data from the `title` attribute to the `aria-label`.

ibacher (approved):

PR #768, opened 2022-01-24
OCLOMRS-1016: Add support for selecting concepts from custom sources to the Bulk Import feature.

hadijahkyampeire:
# JIRA TICKET NAME:
[Add support for selecting concepts from custom sources to the Bulk Import feature](https://issues.openmrs.org/browse/OCLOMRS-1016)

# Summary:
- Implement the showing of all sources when doing bulk import.

# Caveat
- We shall need to do some refactoring and unifying of the code one the other PR with the same work.
- Did not technically use `InfinteScroll` because of its constraints but used some other technique of paginating that is almost like it.


hadijahkyampeire:
@jwnasambu please look at my PR and see what you think about how I load more sources.

hadijahkyampeire:
@suruchee and @ibacher please review my PR for this work. Thanks.

src/apps/dictionaries/pages/AddBulkConceptsPage.tsx
```diff
@@ -22,22 +40,147 @@ const useStyles = makeStyles((theme: Theme) =>
...
+  const sourceUrl = "/sources/";
+  const limit = 20;
+
+  const loadSourcesList = () => {
+    setPaginating(true);
+    setPage(page+1);
+    retrievePublicSources(sourceUrl, queryString, limit, page)
+      .then((res) => {
+        const newList = currentSources?.length  ? [...currentSources, ...res.map((s: any) => ({ name: s.name, sourceUrl: s.url }))] : [...res.map((s: any) => ({ name: s.name, sourceUrl: s.url }))];
+        setCurrentSources(newList);
```
ibacher, line 157:
```suggestion
	const newList = res.map((s: APISource) => ({
	  name: s.name,
	  sourceUrl: s.url
	}));
	setCurrentSources([...(currentSources ?? [], ...newList];)
        setCurrentSources(newList);
```

src/apps/dictionaries/pages/AddBulkConceptsPage.tsx
```diff
@@ -22,22 +40,147 @@ const useStyles = makeStyles((theme: Theme) =>
...
+
+  const [currentSources, setCurrentSources] = useState<
+    { name: string; sourceUrl: string }[]
+  >([]);
+
+  const defaultSources = Object.entries(
+    PREFERRED_SOURCES
+  ).map(([key, value]) => ({ name: key, sourceUrl: value }));
+
+  const allSources = defaultSources.concat(sources.map(s => ({ name: s.name, sourceUrl: s.url })))
```
ibacher, line 128: Since we're re-writing this anonymous function all over the place here, it seems like we should turn it into a named function:

```ts
  s => ({ name: s.name, sourceUrl: s.url })
```

```ts
const mapSources = (s: APISource) => ({
  name: s.name,
  sourceUrl: s.url
})
```

Then replace the other calls with just `mapSources`.

ibacher (commented):
A few thoughts on this, but overall good work @hadijahkyampeire!

hadijahkyampeire:
Thanks, @ibacher for the review I will work on it and get back to you on what I don't understand.

src/apps/dictionaries/pages/AddBulkConceptsPage.tsx
```diff
@@ -22,22 +40,147 @@ const useStyles = makeStyles((theme: Theme) =>
...
+        const newList = currentSources?.length  ? [...currentSources, ...res.map((s: any) => ({ name: s.name, sourceUrl: s.url }))] : [...res.map((s: any) => ({ name: s.name, sourceUrl: s.url }))];
+        setCurrentSources(newList);
+      })
+      .catch((err) => {
+        console.log(err);
+      })
+  };
+
+  useEffect(() => {
+    const defaultSources = Object.entries(
```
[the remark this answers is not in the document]
hadijahkyampeire: True, I didn't see that before.

src/apps/dictionaries/pages/AddBulkConceptsPage.tsx
```diff
@@ -22,22 +40,147 @@ const useStyles = makeStyles((theme: Theme) =>
...
   addConceptsToDictionary: (
     ...args: Parameters<typeof recursivelyAddConceptsToDictionaryAction>
   ) => void;
+  retrievePublicSources: (
+    ...args: Parameters<typeof retrievePublicSourcesAction>
+  ) => Promise<any>;
 }
 
-const AddBulkConceptsPage: React.FC<Props> = ({ addConceptsToDictionary }) => {
+const AddBulkConceptsPage: React.FC<Props> = ({ addConceptsToDictionary, sources = [], retrievePublicSources, meta = {} }) => {
```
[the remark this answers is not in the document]
hadijahkyampeire: @ibacher I am not sure I understand what needs to be done here, For option 2, do you mean making a different component and include all the logic to load sources?

src/apps/dictionaries/pages/AddBulkConceptsPage.tsx
```diff
@@ -22,22 +40,147 @@ const useStyles = makeStyles((theme: Theme) =>
...
+  
+  const { num_found: numberFound = sources.length } = meta;
+
+  const sourceUrl = "/sources/";
+  const limit = 20;
+
+  const loadSourcesList = () => {
+    setPaginating(true);
+    setPage(page+1);
+    retrievePublicSources(sourceUrl, queryString, limit, page)
```
ibacher: Quick thought here: for the full source list, this should actually be `retrievePrivateSources`. The `public` version will only retrieve publicly-visible sources. The `private` version will retrieve all sources including both private and public. Otherwise, we're restricting users to *only* load concepts from sources they've marked as `public`. We should allow users to load concepts from private sources, in case implementers don't wish to publicly publish their source.

hadijahkyampeire: Thanks @ibacher for your review

hadijahkyampeire: @ibacher this has been fixed ☝️

PR #774, opened 2022-03-22
OCLOMRS-1094: Use concept url when removing concept from collection

Piumal1999:
# JIRA TICKET NAME:
[Unable to delete concept from collection with OCL Platform 2.0.117+](https://issues.openmrs.org/browse/OCLOMRS-1094)

# Summary:
Removing a concept within a dictionary (collection) in the Dictionary Manager visually removes the concept, but the concept isn't actually deleted from the collection within OCL and refreshing the page reveals the bug as the concept "reappears" within the dictionary.

As @bmamlin suggested [here](https://issues.openmrs.org/browse/OCLOMRS-1094), I updated the delete request body, but still, the API request is returning a 204 response and fails the deletion. So I'll keep this pull request as a draft for now.

### DIFF
--- a/src/apps/concepts/components/ConceptsActionMenu.tsx
+++ b/src/apps/concepts/components/ConceptsActionMenu.tsx
@@ -34,7 +34,7 @@ const addConceptsToDictionaryMenu = (
 };
 
 const removeConceptsFromDictionaryMenu = (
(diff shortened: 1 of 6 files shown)


suruchee (commented):
Thanks @Piumal1999 for the PR, but we still cannot delete the already existing concepts so let's discuss in squad call to merge it.

